PR #19, opened 2026-05-18
fix: scope remote worktree state by config

therockstorm:
## Summary

Remote worktree state was stored in one user-wide XDG file and only filtered by repository name. This change persists a config namespace for new remote state records, filters remote list/find paths to the active namespace, and keeps unscoped legacy records on disk without exposing them to cleanup or teardown.

## Validation

- `node --run verify`

## Notes

- Claim id: `fnd_sig-feat-cli-command-06d9b3d7d3-_b230ff94a0`
- Legacy remote records without `remoteStateNamespace` are intentionally not returned because their owning config cannot be proven.

Agent session: codex resume 019e3c25-01c5-7ca1-a0ab-3a81e89cc5d3


therockstorm:
@coderabbitai review

PR #25, opened 2026-05-18
feat(config): JSONC user config validated with zod

### LINKED ISSUE #21: Use JSONC config with stable parser/editor support
## Context

PR #15 adds `crew setup discover-repos`, which auto-populates `workspace.knownRepositories` in the user's resolved config:

https://github.com/ClipboardHealth/groundcrew/pull/15

That workflow makes the config file something the CLI needs to read and update, not just something developers import manually. Because of that, a TypeScript config file may not be the best long-term format for this CLI.

## Suggestion

Move the user-facing configuration format toward JSONC instead of TypeScript.

Use a well-tested library such as [`jsonc-parser`](https://www.npmjs.com/package/jsonc-parser) to parse JSONC and apply edits when the CLI needs to update config values such as `workspace.knownRepositories`.

Keep the config schema/types and validation in TypeScript, and validate the parsed config during crew execution.

## Why this matters

A CLI-managed config should be safe and stable for both humans and the application to edit.

JSONC gives us comments and trailing commas while keeping the file as structured data. That makes it easier for the application to parse, validate, migrate, and update the config without executing arbitrary user code or depending on runtime TypeScript loading.

Using `jsonc-parser` specifically helps avoid unstable string manipulation. It can parse JSONC and compute edits while preserving user comments and formatting better than a plain `JSON.parse` / `JSON.stringify` rewrite.

Keeping validation in the crew execution path is still important. The config file format should be easy to manipulate, but execution should continue to enforce the expected shape and produce actionable errors when the file is invalid.

## Acceptance criteria

- User-facing config can be represented as JSONC.
- Config parsing uses a maintained JSONC parser/editor library such as `jsonc-parser`.
- CLI-driven config updates avoid ad hoc string manipulation.
- Existing config validation remains in the crew execution path.
- PR #15's config update behavior is aligned with this approach or has a clear migration path.

pgbezerra:
## Summary

Closes #21.

- Switch the user-facing config from a TypeScript module to JSONC, parsed with [`jsonc-parser`](https://www.npmjs.com/package/jsonc-parser). Config is now structured data; comments and trailing commas round-trip.
- Replace the ad-hoc per-field validators in `src/lib/config.ts` with a single [zod](https://zod.dev/) schema. Validation still runs on every `loadConfig()` call, so the crew execution path enforces the same shape it always did — just with path-prefixed messages that flow from one source of truth.
- Ship `configExample.jsonc` (replaces `configExample.ts`). README's quickstart copies the new file. A `config:check` step in `node --run verify` re-validates the published example against the live schema on every push, so the file and the schema cannot drift.
- Ship `scripts/migrateConfigToJsonc.mts` plus a `config:migrate` npm script. It strips the TS wrapper, quotes bare keys, rewrites single-quoted strings and numeric separators, and preserves every comment + trailing comma — the conversion path the issue explicitly asks for. The legacy-config error surfaced by the CLI tells the user exactly which command to run.

### Alignment with PR #15 (`crew setup discover-repos`)

PR #15 is still open against `main` and rewrites `workspace.knownRepositories` directly inside `config.ts`. Once this PR lands, that flow needs to swap its string-manipulation step for `jsonc-parser`'s `modify` + `applyEdits` against the JSONC file — same target field, different writer. The migration path is the converter script in this PR; existing users run it once, and `crew setup discover-repos` keeps its idempotent semantics on top.


therockstorm:
I'd rather use https://github.com/cosmiconfig/cosmiconfig to support various config formats.

PR #29, opened 2026-05-19
feat: support Linux local runs with a Bubblewrap isolation backend

### LINKED ISSUE #27: Support Linux local runs with a Bubblewrap isolation backend
## Problem

Groundcrew currently treats local execution as macOS + Safehouse only. On Linux/WSL, users must label tickets with `agent-remote` and run through the remote runner, even though the workspace layer already supports Linux-capable backends such as tmux/cmux.

We should add a Linux local isolation model that is as close as practical to Safehouse's workflow: low-friction local CLI execution, deny-first filesystem access, explicit path grants, and compatibility with normal coding-agent subprocess trees.

## Recommendation

Implement Bubblewrap (`bwrap`) as the default Linux local runner backend, while still allowing an explicit `none` runner for users who intentionally want unsandboxed local execution.

Bubblewrap is a good fit because it:

- runs ordinary CLI commands locally without a Docker/Podman daemon or image workflow;
- creates Linux namespaces around the agent process;
- lets Groundcrew construct an explicit filesystem view with read-only and read-write bind mounts;
- supports private `/tmp`, `/proc`, process/session isolation, and optional network namespace behavior;
- is available in most Linux distribution package repositories.

This should be implemented as a Groundcrew-owned policy compiler rather than exposing raw `bwrap` args, because unsafe bind mounts such as `/` would defeat the sandbox.

## Proposed shape

Add an explicit local runner strategy, for example:

```ts
local: {
  runner: "safehouse" | "bubblewrap" | "none";
  linux?: {
    allowedReadPaths: string[];
    allowedWritePaths: string[];
    envPass: string[];
    network: "host" | "none";
  };
}
```

Default behavior:

- macOS defaults to `safehouse`.
- Linux defaults to `bubblewrap`.
- `none` remains available as an explicit unsafe escape hatch, with strong config validation and doctor warnings. It must not be the implicit Linux default.

## Initial Bubblewrap policy

Suggested default grants:

- worktree: read-write;
- Groundcrew prompt/temp directory: only what launch cleanup requires;
- `/tmp`: private tmpfs;
- `/proc`: new proc namespace;
- `/dev`: minimal device set;
- system/runtime/toolchain paths: read-only, e.g. `/usr`, `/bin`, `/lib`, `/lib64`, `/etc/ssl/certs`, `/etc/resolv.conf`;
- home directory: not mounted wholesale;
- model-specific auth/config dirs only when needed, e.g. `~/.codex`, `~/.claude`, `~/.config/gh`, `~/.gitconfig`, `~/.config/git`;
- env passthrough: explicit allowlist only.

## Implementation notes

- Split local command wrapping in `buildLaunchCommand()` so Safehouse, Bubblewrap, and `none` are separate launch backends.
- Call `ensureClearance()` only when the selected local runner is Safehouse.
- Update host detection and `crew doctor` so Linux reports Bubblewrap readiness by default instead of recommending `agent-remote`.
- If `local.runner = "none"`, doctor should make the unsandboxed behavior explicit and local setup should never silently select it.
- Add focused tests for generated `bwrap` args, especially to prevent accidental broad mounts of `$HOME`, `/`, SSH keys, browser profiles, or broad credential directories.
- Keep the config/policy surface compatible with a future Landlock hardening layer.

## Alternatives considered

- **Rootless Docker/Podman:** stronger packaging boundary, but worse fit for Safehouse-like native local CLI execution. Requires image/runtime setup and introduces bind mount/UID mapping complexity.
- **Firejail:** easy command wrapper, but SUID/profile behavior varies by distro and is less aligned with Groundcrew owning a predictable policy compiler.
- **Landlock:** conceptually close to Safehouse's path access model and worth considering later, but it would require a custom wrapper/dependency and mostly covers filesystem restrictions rather than the full namespace story.

## Acceptance criteria

- Linux users can run a ticket locally with the default Linux local runner, `bubblewrap`, and no `agent-remote` label.
- Linux users may explicitly configure `local.runner = "none"`, but Groundcrew never selects `none` implicitly.
- `crew doctor` reports Bubblewrap local-runner status on Linux by default.
- `crew doctor` clearly warns when `local.runner = "none"` is configured.
- Local Linux Bubblewrap runs do not mount `$HOME` or `/` broadly by default.
- Agent commands still receive the staged prompt and run from the ticket worktree.
- Existing macOS Safehouse local behavior remains unchanged.

References:

- Safehouse overview: https://agent-safehouse.dev/docs/overview
- Safehouse isolation comparison: https://agent-safehouse.dev/docs/isolation-models
- Bubblewrap README: https://github.com/containers/bubblewrap/blob/main/README.md
- Linux Landlock docs: https://docs.kernel.org/userspace-api/landlock.html

pgbezerra:
## Summary

- Add `local.runner` config (`auto` | `safehouse` | `bubblewrap` | `none`) so Linux/WSL hosts have a first-class local backend instead of being routed to `agent-remote`. `auto` resolves to `safehouse` on macOS and `bubblewrap` on Linux; `none` is an explicit unsafe escape hatch that is never selected implicitly.
- Implement Bubblewrap as a Groundcrew-owned policy compiler: deny-first filesystem view, private `/tmp`/`/proc`, namespace isolation, env passthrough allowlist (`--clearenv` + `--setenv`), and broad-mount rejection (`/`, `$HOME`, `/etc`, `/home`, …) so a config typo can't defeat the sandbox.
- Route HTTP egress through the same `clearance` proxy as macOS Safehouse — `HTTP_PROXY`/`HTTPS_PROXY`/`ALL_PROXY` (+ lowercase) auto-injected into the bwrap env, `NO_PROXY=localhost,127.0.0.1,::1`. `clearance-allow-hosts` is the single source of truth for hostname egress on both platforms.
- Split local command wrapping in `buildLaunchCommand` into three backends; gate `ensureClearance` so it runs for `safehouse` and `bubblewrap` but not `none`.
- Surface the resolved runner in `crew doctor`, with install hints for missing `bwrap` and a loud warning when `local.runner='none'` is set.

## Known limitation: Linux egress is honor-system

The clearance proxy is fully enforced for code that honors `HTTP_PROXY` (claude / codex / curl / git / npm / etc.), so the typical attack surface matches macOS Safehouse. **But** bwrap with `network: "host"` shares the host's net namespace, so a hostile or buggy process inside the sandbox can ignore `HTTP_PROXY` and connect directly to the internet. macOS Safehouse closes this via `clearance-only.sb` denying all `network-outbound` at the sandbox layer; Linux has no equivalent without root-level nftables/iptables or a routed namespace via `slirp4netns` — out of scope for this PR. `network: "none"` shuts egress entirely (at the cost of no API access).

Verified locally: allowed hosts (`api.anthropic.com`) reach through the proxy and return; disallowed hosts (`example.com`) are blocked by clearance; direct connection bypass (`curl --noproxy "*" 1.1.1.1`) succeeds — matching the limitation above.

Closes #27


pgbezerra:
Tomorrow I'll reintroduce `docker sdx` in a fresh PR.

CC: @therockstorm @jamie-tm

PR #33, opened 2026-05-19
feat: support GROUNDCREW_LINEAR_API_KEY env var

paulbaranowski:
## Summary

The Linear API key was previously read only from `LINEAR_API_KEY`, which collides with other tools that consume the same name (shell rc files, direnv, IDE extensions). This PR adds `GROUNDCREW_LINEAR_API_KEY` as a higher-priority alternative so users can scope a Linear workspace specifically for `groundcrew` without disturbing other tooling.

Resolution order: `GROUNDCREW_LINEAR_API_KEY` first, then `LINEAR_API_KEY` (empty strings count as unset). Existing setups using `LINEAR_API_KEY` continue to work unchanged. `crew doctor` now reports which variable supplied the key and lists both names when neither is set. The "no Linear project found" error message no longer hardcodes a specific env var name.

## Validation

- `node --run verify` — 19 test files, 538 tests passing, 100% coverage on statements/branches/functions/lines, 0 lint/format errors.
- TDD red-green: 9 new failing tests in `util.test.ts` + 3 in `doctor.test.ts` exercising the full resolution matrix (only-legacy, only-prefixed, both-set precedence, both-empty, fallback-from-empty) before any implementation; all green after.

## Notes

- Doctor output shape changes from `[ok] $LINEAR_API_KEY  — set` to `[ok] linear api key  — set via $GROUNDCREW_LINEAR_API_KEY` (or `$LINEAR_API_KEY`). Any user script that grepped the prior wording will need adjustment; impact is expected to be low.
- Design spec preserved at `docs/superpowers/specs/2026-05-19-groundcrew-linear-api-key-env-var-design.md` for reviewer reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,17 +49,19 @@ This installs the `crew` binary. `@clipboard-health/clearance` is pulled in tran
 
    `crew` resolves the config path as: `GROUNDCREW_CONFIG` if set → `${XDG_CONFIG_HOME:-$HOME/.config}/groundcrew/config.ts` if it exists → a `config.ts` sitting next to `crew`'s own source files (only useful from a local checkout; see [Hacking on groundcrew](#hacking-on-groundcrew)). Set `GROUNDCREW_CONFIG` only when you want to override the XDG location.
 
-4. **Provide a Linear API key.** `crew` expects `LINEAR_API_KEY` in its environment. Any mechanism works — shell export, [direnv](https://direnv.net/), a `.env` file you `source`, or piping through `op run` if you store the credential in 1Password:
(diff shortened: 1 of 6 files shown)


pgbezerra (commented):
Hello! Thanks for your contribution!

Would you mind removing the superpower plan?

paulbaranowski:
> Hello! Thanks for your contribution!
> 
> Would you mind removing the superpower plan?

Oops, didnt mean to include that! will remove

pgbezerra (approved):

PR #42, opened 2026-05-19
feat(sdx): opt-in SSH agent forwarding into sandbox

pgbezerra:
## Summary

- Adds `sandbox.forwardSshAgent` config knob for the sdx runner so the agent CLI inside the sandbox can `git push` over SSH using the host's SSH agent (e.g. 1Password).
- Opt-in: existing sdx configs are unaffected. When enabled but `SSH_AUTH_SOCK` is unset at run time, groundcrew logs a warning and skips the forward — the launch still proceeds.
- Mechanics: at `sbx create` the directory holding `$SSH_AUTH_SOCK` is bind-mounted into the sandbox; on every `sbx exec` `-e SSH_AUTH_SOCK` passes the socket path through to the agent process.


pgbezerra:
I'm rethinking this sdx ssh agent forward, and maybe we shouldn't do it. maybe we should actually commit to git without signing the commits, anyway the commits will be made by robot, you know? And crew should have a `sandbox` command to configure sdx, like regenerating the sandbox for each model, configuring, managing our template.

PR #47, opened 2026-05-19
feat(workspaces): add herdr backend alongside cmux and tmux

pgbezerra:
## Summary

- Adds a third workspace backend, [`herdr`](https://herdr.dev), opt-in via `workspaceKind: "herdr"` and auto-selected ahead of `tmux` when the binary is on PATH.
- Refactors `src/lib/workspaces.ts` into an adapter registry where each adapter self-reports `kind`, `autoPriority`, and `isAvailable(host)`. A fourth backend lands as one adapter object instead of three coordinated edits across the resolver, the auto-fallback chain, and the capability map.
- Introduces `WorkspaceStatus.state` (`idle | working | blocked | unknown`) so adapters with a real agent-state API (herdr's `pane.report_agent`) can surface it cleanly; cmux/tmux silently drop the field.

## Why

Herdr is appealing for two structural reasons that cmux can't match without the workarounds we've accumulated:

1. **Native agent-state API.** Herdr ships `pane.report_agent` with a typed semantic state and renders pane visuals from it. cmux v2+ dropped `set-status` entirely — we've been swallowing the resulting "unknown command" errors. Herdr lets us paint status cleanly with no special-casing.
2. **No client/server host split.** cmux's macOS desktop app drives PTYs on remote hosts, which forced the `probeCurrentCmuxRemote` + `buildSshWrappedCommand` workaround in `workspaces.ts` so a remote-launched workspace lands on the right host. Herdr's server colocates with the PTYs, so this entire class of problem doesn't exist for the herdr adapter.

Adding the backend forced the resolver refactor, which also pays back the growing `if/else` chain the existing two backends had accumulated.

## Caveats — why draft

A handful of herdr CLI shapes are inferred from the public docs at [herdr.dev/docs/cli-reference](https://herdr.dev/docs/cli-reference/) without a binary to validate against locally:

- `herdr workspace create` stdout format (JSON envelope vs. bare id)
- The `--json` flags on `herdr workspace list` / `herdr pane list` and their envelope shape (top-level array vs. `{ workspaces: [...] }` / `{ panes: [...] }`)
- Whether `workspace create` provisions a default pane (assumed yes, and the adapter picks the first one from `pane list`)

The parse helpers accept both shapes, fall back when needed, and are covered at 100%. Before un-drafting, those three flag shapes should be checked against a real `herdr` install — the relevant code is `parseHerdrWorkspaceList`, `parseHerdrPaneList`, and `extractHerdrWorkspaceId` in `src/lib/workspaces.ts`.


pgbezerra:
Bad experience just as cmux. Worth the shot, but I'll not continue it.

PR #49, opened 2026-05-20
feat(prompts): ship per-model default prompts as Markdown files

paulbaranowski:
## Summary

- Replace the 5-line minimal default prompt with opinionated per-model defaults (Claude, Codex, generic).
- Prompts live in [`prompts/*.md`](./prompts) at the package root — editable as plain Markdown, linked from the README, included in `npm publish` via the `files` array.
- New resolution chain: user override (`prompts.initial`) → per-model default → generic fallback.
- All three shipped defaults instruct the agent to stay autonomous, run tests, spawn a sub-agent review pass, open a **draft** PR, and transition the Linear ticket to **"In Review"** before stopping.

## Why

The shipped default was a 5-line template ("Begin work on X, here's the description") that left every workflow decision to the agent. In practice agents stopped to ask design questions mid-task, never opened drafts, and didn't move the ticket out of "In Progress." Replacing the default with opinionated per-agent prompts makes the out-of-box experience match what most groundcrew users will configure manually anyway.

## What changed

- **New:** [`prompts/claude.md`](./prompts/claude.md), [`prompts/codex.md`](./prompts/codex.md), [`prompts/generic.md`](./prompts/generic.md). Read these to see exactly what each agent receives.
- **New:** `src/lib/prompts.ts` — loads the `.md` files at module-init via `readFileSync`, exposes `resolvePromptForModel(config, modelName)` plus the three prompt constants and `DEFAULT_PROMPTS_BY_MODEL`.
- **Modified:** `ResolvedConfig.prompts.initial` widened from `string` to `string | undefined`. User overrides still work as before; the difference is the built-in default is no longer baked into this field.
- **Modified:** `src/commands/setupWorkspace.ts` — `stagePrompt` now routes through `resolvePromptForModel(config, model)`.
- **Modified:** `package.json` adds `prompts` to the `files` array so the directory ships with `npm publish` (verified via `npm pack --dry-run`).
- **Modified:** `.jscpd.json` ignores `prompts/**` — the three Markdown files intentionally share ~90% of their content as readable, standalone artifacts.
- **Modified:** README config-reference row and gotcha now link to the prompt files. `configExample.ts` comment points readers to the files, not the constants.

## Decisions

- **Per-model files, not one shared file with a templating system.** Each prompt is supposed to be a self-contained artifact a human can read top-to-bottom and understand exactly what their agent sees. Templates would save ~50 lines of repeated text but defeat that property. Trade-off acknowledged via the jscpd ignore entry.
- **\"In Review\" hard-coded in all three prompts.** It's the single team-specific assumption baked into the defaults. Teams using a different status name override \`prompts.initial\`. Adding a placeholder for the status name was considered but rejected to keep the prompt-placeholder allowlist at four (the existing four are universally available; a fifth would be sometimes-undefined).
- **\`superpowers:using-superpowers\` only in the Claude prompt.** Claude Code is the only agent with Claude Code skills, and the \`superpowers\` plugin is the load-bearing discipline overlay for Claude Code workflows. Listed in the Claude prompt with no \"if available\" qualifier — if the plugin isn't installed, Claude Code reports the skill as missing and continues, which is the right failure mode (loud, recoverable).
- **Markdown files, not embedded strings.** Easier to read on GitHub, easier to diff when the workflow evolves, lets the README link directly to the prompt instead of referencing TypeScript constants users don't recognize.

## Test plan

- [x] \`node --run verify\` — all 10 checks pass (architecture, build, cpd, format, knip, lint, markdown, spell, syncpack, test)
- [x] 701 tests passing, 100% coverage maintained
- [x] \`npm pack --dry-run\` confirms \`prompts/*.md\` ship in the package (~5.3KB total)
- [x] Existing \`prompts.initial\` user override tests in \`config.test.ts\` unchanged and still passing
- [x] New \`src/lib/prompts.test.ts\` covers all four resolution paths (user override; per-model default for claude, codex; generic fallback for unknown models) and content invariants (all four placeholders present, no unknown placeholders, draft + In Review mentioned, superpowers reference only in Claude)

## Out of scope

- Per-model user overrides (e.g. \`models.definitions.claude.prompt\`). Not requested; existing \`prompts.initial\` global-override covers the common case.
- Parameterizing \"In Review\" status name. Could be added later if multiple teams need it; the placeholder allowlist would need an addition.


therockstorm:
Thanks for the PR. I understand the underlying problem of the 5-line default leaving too much workflow up to the agent, and the prompt content you wrote is solid. But I don't think the shipped package default is the right place for it.

The three files aren't really three prompts. `codex.md` and `generic.md` are identical. `claude.md` differs from them by two sentences: the superpowers:using-superpowers invocation and the "no Generated with Claude Code footer" line.

More importantly, the content is user-specific, not model-specific e.g. superpowers:using-superpowers assumes a specific plugin. Shipping these as the default means every external groundcrew user gets a config that doesn't match their workflow.

I like the generic mechanism we already have: Your gitignored config.ts at the package root can do:

```ts
prompts: {
  initial: readFileSync("./prompts/claude.md", "utf8"),
}
```

Keep the prompt file next to your config, versioned in your own dotfiles and the package stays opinion-free.

Pushback welcome if I'm missing something.

paulbaranowski:
> I like the generic mechanism we already have: Your gitignored config.ts at the package root can do:
> 
> ```ts
> prompts: {
>   initial: readFileSync("./prompts/claude.md", "utf8"),
> }
> ```
> 
> Keep the prompt file next to your config, versioned in your own dotfiles and the package stays opinion-free.
> 
> Pushback welcome if I'm missing something.

It took me a while to figure out I needed to do this, and then it took a while to figure out what needed to be in the prompt to actually make it work. It wouldnt be obvious especially to someone just dabbling and trying out the project. If you dont include a default, every user has to re-discover how to construct this prompt. There should be some sensible default so that someone can download and just try it out, and it works to completion instead of immediately hanging. It's fine if we narrow it down to one prompt and strip out the plugin suggestions (the agent will just skip if they are not installed). Also, I just discovered the babysit-pr skill which should be integrated into the prompt as well. 

I think it would make sense to have the basic prompt by default + an advanced prompt ready to go for those people who have everything: all installed plugins + coderabbit.

therockstorm:
@paulbaranowski How about https://github.com/ClipboardHealth/groundcrew/pull/63?

No every repo is going to have /babysit-pr. I don't want to force a workflow on people.

PR #57, opened 2026-05-20
docs(readme): document build-time secret-passing model (tg-3691)

Davidster:
## Summary

- Adds a **Secrets** section to the README (between Quickstart and Runners) so new users can understand how `NPM_TOKEN` / `BUF_TOKEN` flow through groundcrew without grepping `src/lib/launchCommand.ts` or `src/lib/buildSecrets.ts`.
- Covers: recognized var names (with link to `BUILD_SECRET_NAMES`), `secrets.env` staging at mode `0600`, source-before-setup, `unset` + `rm -rf` of the prompt dir before `exec`'ing the agent — net effect: by the time the agent runs, the values are gone from the environment and the file is gone from disk.

Needs to land in **ClipboardHealth/groundcrew**.

Closes tg-3691.

## Test plan

- [x] `node --run verify` (685 tests, 100% coverage)
- [x] Reviewer skim: confirm the Secrets section reads end-to-end without needing the source

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,25 @@ Installs the `crew` binary. `@clipboard-health/clearance` is pulled in transitiv
    crew run --watch               # poll forever
    ```
 
+## Secrets
+
+Groundcrew forwards a small allowlist of build-time secrets from your shell into the setup phase (so `npm install` can authenticate against private registries) and then strips them before the agent runs. The agent process never inherits these values in its environment.
+
+**Recognized names.** Defined in [`BUILD_SECRET_NAMES`](./src/lib/buildSecrets.ts):
+
+- `NPM_TOKEN`
+- `BUF_TOKEN`
+
+Set them in the shell you run `crew` from. Anything not in this list is ignored by the secret-shuttling path.
+
+**Flow.** For each ticket:
+
+1. If any recognized var is set and non-empty, groundcrew writes `secrets.env` (mode `0600`) into the ticket's temp prompt dir as `KEY='value'` lines — see `stageBuildSecrets` in [`src/commands/setupWorkspace.ts`](./src/commands/setupWorkspace.ts).
+2. The launch script sources `secrets.env` with `set -a` so the values are exported into the setup phase only (and under `sdx`, forwarded into the sandbox via `-e KEY` flags).
+3. After setup completes, the script `unset`s every name in `BUILD_SECRET_NAMES` before `exec`'ing the agent. See `sourceSecretsLine` / `unsetSecretsLine` in [`src/lib/launchCommand.ts`](./src/lib/launchCommand.ts).
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -105,6 +105,25 @@ Installs the `crew` binary. `@clipboard-health/clearance` is pulled in transitiv
...
+
+Set them in the shell you run `crew` from. Anything not in this list is ignored by the secret-shuttling path.
+
+**Flow.** For each ticket:
+
+1. If any recognized var is set and non-empty, groundcrew writes `secrets.env` (mode `0600`) into the ticket's temp prompt dir as `KEY='value'` lines — see `stageBuildSecrets` in [`src/commands/setupWorkspace.ts`](./src/commands/setupWorkspace.ts).
+2. The launch script sources `secrets.env` with `set -a` so the values are exported into the setup phase only (and under `sdx`, forwarded into the sandbox via `-e KEY` flags).
+3. After setup completes, the script `unset`s every name in `BUILD_SECRET_NAMES` before `exec`'ing the agent. See `sourceSecretsLine` / `unsetSecretsLine` in [`src/lib/launchCommand.ts`](./src/lib/launchCommand.ts).
+
+**Caveat.** `secrets.env` itself stays on disk inside the prompt dir for the agent's lifetime — the launch command `rm -rf`s the temp dir on exit, but a sufficiently-motivated agent process could read the file while it's running, even though the values are no longer in its environment. If that's in your threat model, don't set these vars when invoking `crew`.
```
Davidster: this caveat is not correct. see this:

```
Investigation: no action needed.

secrets.env is already removed before the agent process starts — the ticket's premise that "the file itself is never explicitly removed"

is incorrect.

Lifecycle:

Created in setupWorkspace.ts:79-80 (mode 0600) under the temp prompt dir.

Sourced (set -a && . file && set +a) by the launch script before setup.

NPM_TOKEN/BUF_TOKEN are unset after setup completes.

The whole prompt dir is rm -rf'd (launchCommand.ts:140 host, :208 sdx) before exec <agent>, so secrets.env is gone from disk and the

env vars are gone from the shell when the agent execs.

The rollback path (setupWorkspace.ts:284) also wipes the prompt dir if setup throws.

The agent process never inherits the file or the secrets. Lifetime is bounded to the launch shell, by design — and stageBuildSecrets's

JSDoc already documents this ("The temp dir is rm -rf'd by the launch command (and rollback path), so cleanup is already handled.").

Closing — no code or doc changes needed. If we later want a tighter window (delete right after sourceSecretsLine instead of after setup

completes), that's a separate small change worth a paragraph of rationale, but the current model is sound.
```

from TG-3692

Davidster: Good catch — you're right, the prompt dir (and `secrets.env` with it) is `rm -rf`'d before `exec`'ing the agent, so the file doesn't survive into the agent process either. Rewrote the third bullet to say so explicitly and dropped the misleading caveat in fa15df2.

therockstorm (approved):

PR #59, opened 2026-05-20
fix(doctor): fail when codexbar is missing for a usage-gated model

Davidster:
Lands in `ClipboardHealth/groundcrew`. Closes TG-3689.

## Summary
- The default `claude` entry in `DEFAULT_MODEL_DEFINITIONS` carries a `usage` block, so on a host without codexbar `crew run` skips every claude-bound ticket at dispatch time with only an opaque log line (`claude session at Infinity% (> 85%) — skipping its tickets`).
- `crew doctor` reported codexbar as `[? ] optional`, so the problem never surfaced before dispatch.
- Doctor now treats codexbar as required whenever any enabled model has `usage` configured. The failure hint names the gated model(s) and points at both remediations: install codexbar, or override `models.definitions.<name>.usage` to disable gating.
- Pulled the model-iteration helper out of `src/lib/usage.ts` as an exported `gatedModels` so doctor and the runtime usage path share one source of truth.

## Test plan
- [x] `node --run verify` (686 tests, 100% coverage)
- [ ] `crew doctor` without codexbar on PATH → fails with the new hint
- [ ] `crew doctor` with codexbar installed → `[ok] codexbar`
- [ ] Override the default `claude.usage` to disable gating → codexbar check is skipped

### DIFF
--- a/src/commands/doctor.test.ts
+++ b/src/commands/doctor.test.ts
@@ -820,10 +820,9 @@ describe(doctor, () => {
     const actual = await doctor();
 
     expect(actual).toBe(false);
-    expect(consoleLog.output()).toContain("[--] codexbar");
-    expect(consoleLog.output()).toContain("required for usage gating on `claude`");
-    expect(consoleLog.output()).toContain("install codexbar");
-    expect(consoleLog.output()).toContain("disable gating");
+    expect(consoleLog.output()).toContain(
+      "[--] codexbar  — required for usage gating on `claude` — install codexbar, or set `models.definitions.<name>.usage` to disable gating",
+    );
   });
 
   it("reports codexbar as ok when usage is configured and codexbar is on PATH", async () => {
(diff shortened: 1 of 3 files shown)


src/commands/doctor.test.ts
```diff
@@ -819,10 +819,33 @@ describe(doctor, () => {
 
     const actual = await doctor();
 
-    // codexbar is not required; doctor still passes when only it is missing.
+    expect(actual).toBe(false);
+    expect(consoleLog.output()).toContain("[--] codexbar");
```
Davidster: why not just assert the whole console output with a single (template?) string?

Davidster: Good call — collapsed into a single full-line `toContain` in 3002c2e.

therockstorm (approved):

PR #61, opened 2026-05-20
feat(doctor): warn when linear.statuses.todo missing on project teams

Davidster:
## Summary
- `crew run --watch` filters Linear tickets server-side by `linear.statuses.todo`. When the configured status name doesn't exist on a project team, dispatch silently produces "no eligible tickets" with no hint to the user.
- Reproducing case: Team Gaia at Clipboard Health has no `Todo` status — `Backlog` plays that role — so a default groundcrew config targeting a Gaia project sees zero dispatches.
- `crew doctor` now fetches each team attached to the configured Linear project and fails with a `[--]` line listing that team's actual statuses when `linear.statuses.todo` is missing, so the user can remap it.

Needs to land in `ClipboardHealth/groundcrew`.

Closes TG-3688.

## Test plan
- [x] `node --run verify` (all checks pass, 100% coverage)
- [ ] On a Linear project whose team lacks `Todo`, `crew doctor` exits non-zero and prints the available statuses with a remap hint
- [ ] On a project whose team(s) all have `Todo`, `crew doctor` reports `[ok]` for each team
- [ ] When the Linear API key is unset, the team-status check is skipped (no double-warning)
- [ ] When the team-status GraphQL query fails, doctor surfaces a non-blocking `[? ]` warning rather than aborting

### DIFF
--- a/src/commands/doctor.test.ts
+++ b/src/commands/doctor.test.ts
@@ -190,13 +190,24 @@ function activeNodes(count: number): { id: string }[] {
   return Array.from({ length: count }, (_value, index) => ({ id: `active-${index}` }));
 }
 
+interface TeamStub {
+  key: string;
+  name: string;
+  statuses: string[];
+}
+
+const DEFAULT_PROJECT_TEAMS: TeamStub[] = [
+  { key: "TEAM", name: "Team Default", statuses: ["Todo", "In Progress", "Done"] },
+];
+
 function makeLinearClient(
   options: {
     issue?: RawIssueStub | null;
     activePages?: number[];
+    projectTeams?: TeamStub[] | "no-project" | null;
   } = {},
 ): LinearClient {
-  const { issue = rawIssue(), activePages = [0] } = options;
+  const { issue = rawIssue(), activePages = [0], projectTeams = DEFAULT_PROJECT_TEAMS } = options;
   let activePageIndex = 0;
   let blockerPageIndex = 0;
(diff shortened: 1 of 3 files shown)


src/commands/doctor.test.ts
```diff
@@ -909,6 +945,81 @@ describe(doctor, () => {
...
+      makeLinearClient({
+        projectTeams: [{ key: "GAIA", name: "Gaia", statuses: ["Backlog", "In Progress", "Done"] }],
+      }),
+    );
+
+    const actual = await doctor();
+
+    expect(actual).toBe(false);
+    const output = consoleLog.output();
+    expect(output).toContain('[--] linear.statuses.todo "Todo" on team GAIA');
```
Davidster: can we assert the whole doctor output line in one assert? would be clearer to see that it outputs a well formatted log.

src/commands/doctor.test.ts
```diff
@@ -909,6 +945,81 @@ describe(doctor, () => {
     expect(checkedCommands()).not.toContain("cmux");
   });
 
+  it("fails when linear.statuses.todo is missing on a project team", async () => {
```
Davidster, line 948: do we have a test that checks that the doctor succeeds when the team doesn't have a 'todo' status but the crew config provides a proper mapping to mitigate?

Davidster:
Closing for now — keeping TG-3688 open to revisit later.

PR #71, opened 2026-05-21
feat(workflow): add interrupt command (ORB-2185)

sarthak-cbh:
## Summary

Project: [Groundcrew Daily Workflow Improvements](https://linear.app/clipboardhealth/project/groundcrew-daily-workflow-improvements-5d9dfbed8c9d)

This PR adds `crew interrupt <ticket> [--reason <text>]`. It stops the live workspace when present, preserves the worktree, and records the run as interrupted.

## Contribution to the Stack

This adds the safe pause operation. It depends on persisted run state from #70 and enables #72 to resume preserved work with context.

Overall stack theme: make Groundcrew runs easier to recover, pause, inspect, resume, and clean up during daily ticket work.

## Stack Order

1. #69 shared launch plumbing
2. #70 persisted run state
3. #71 interrupt command — current PR
4. #72 resume command
5. #73 doctor recovery state
6. #74 recovery workflow docs

## Validation

- `node --run verify` passed locally on this branch.
- `node --run verify` passed on every branch in the stack after restacking.
- Pre-push verification passed before pushing the rewritten stack.
- GitHub `main` workflow passed for this PR.
- Black-box top-of-stack flow exercised `crew interrupt` against a real tmux-backed local run.

## Linear

[ORB-2185](https://linear.app/clipboardhealth/issue/ORB-2185/add-crew-interrupt-to-pause-a-run-without-deleting-work)

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,14 +316,16 @@ crew run                                                 # one-shot dispatch
 crew run --watch                                         # poll forever
 crew run --ticket <TICKET>                               # provision one ticket and exit
 crew setup repos [--dry-run] [<repo>...]
+crew interrupt <TICKET> [--reason <text>]                # stop the live workspace, keep the worktree
+crew resume <TICKET>                                     # reopen an existing ticket worktree
 crew cleanup <TICKET>                                    # tear down every worktree carrying this ticket
 ```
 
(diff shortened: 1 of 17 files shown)


therockstorm:
I’d revisit the `workspaces.interrupt` success semantics around the cmux close path.

`interruptWorkspace` currently returns `{ kind: "interrupted" }` after `workspaces.close(config, name)` resolves. For cmux, `close` first lists to find the workspace id; if that second `list-workspaces` call fails, it logs and returns without closing anything. That means `crew interrupt` can record run state as `interrupted` even though the close operation was skipped because cmux became unavailable between the probe and close.

This is lower probability than the resume case, but it makes the lifecycle state overconfident. I’d make the close path expose whether it actually closed/confirmed absence, or have `workspaces.interrupt` perform the id-based close directly so `unavailable` stays `unavailable` instead of becoming `interrupted`.

sarthak-cbh:
Addressed. `workspaces.close` now returns an explicit `closed` / `missing` / `unavailable` outcome, cmux close reports `unavailable` when it cannot confirm the close, and `interrupt` only records interrupted after a confirmed close or confirmed missing workspace. Verified with `node --run verify` plus the isolated CLI flow.

therockstorm (approved):

PR #72, opened 2026-05-21
feat(workflow): add resume command (ORB-2186)

sarthak-cbh:
## Summary

Project: [Groundcrew Daily Workflow Improvements](https://linear.app/clipboardhealth/project/groundcrew-daily-workflow-improvements-5d9dfbed8c9d)

This PR adds `crew resume <ticket>`. It reopens the existing ticket worktree, stages a continuation prompt, includes prior interrupt context when available, and updates resume count in run state.

## Contribution to the Stack

This adds the continuation step after `interrupt`. It uses shared launch plumbing from #69 and run state from #70/#71 so resumed work does not need to rediscover context from scratch.

Overall stack theme: make Groundcrew runs easier to recover, pause, inspect, resume, and clean up during daily ticket work.

## Stack Order

1. #69 shared launch plumbing
2. #70 persisted run state
3. #71 interrupt command
4. #72 resume command — current PR
5. #73 doctor recovery state
6. #74 recovery workflow docs

## Validation

- `node --run verify` passed locally on this branch.
- `node --run verify` passed on every branch in the stack after restacking.
- Pre-push verification passed before pushing the rewritten stack.
- GitHub `main` workflow passed for this PR.
- Black-box top-of-stack flow exercised repeated `crew resume` against a real tmux-backed local run.

## Linear

[ORB-2186](https://linear.app/clipboardhealth/issue/ORB-2186/add-crew-resume-with-continuation-context)

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,14 +316,16 @@ crew run                                                 # one-shot dispatch
 crew run --watch                                         # poll forever
 crew run --ticket <TICKET>                               # provision one ticket and exit
 crew setup repos [--dry-run] [<repo>...]
+crew interrupt <TICKET> [--reason <text>]                # stop the live workspace, keep the worktree
+crew resume <TICKET>                                     # reopen an existing ticket worktree
 crew cleanup <TICKET>                                    # tear down every worktree carrying this ticket
 ```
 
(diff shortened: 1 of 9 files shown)


therockstorm:
`resumeWorkspace` should fail closed when `workspaces.probe(config)` returns `{ kind: "unavailable" }`.

Right now `failIfWorkspaceAlreadyLive` only blocks when the probe is `ok` and contains the ticket, so an unavailable cmux/tmux probe is treated like “no live workspace.” That is risky for resume: if the backend cannot list sessions, `crew resume <ticket>` can open a duplicate workspace for a run that is already live. This conflicts with the `WorkspaceProbe` contract in `workspaces.ts`, which says `unavailable` means “we don’t know,” not “empty.”

Suggested behavior: if `probe.kind === "unavailable"`, throw an error that says Groundcrew could not verify whether the workspace is already live and ask the user to retry/inspect the backend manually. Keep the existing “already live” error for the positive match case.

sarthak-cbh:
Addressed. `resume` now fails closed when workspace probing is unavailable, including backend error details, before opening a workspace or writing run state. Added probe-unavailable coverage and verified with `node --run verify` plus the isolated CLI flow.

therockstorm (approved):

PR #80, opened 2026-05-22
fix(dispatch): surface parent-ticket skips in doctor and daemon log

paulbaranowski:
## Summary

Symptom: `crew doctor --ticket <id>` reports "→ would be dispatched on next tick" for a Todo ticket, but the daemon never picks it up and just logs "No Todo tickets to pick up" each tick.

Root cause: `boardSource.fetchBoard` silently drops parent tickets (issues with sub-issues) via the client-side filter `node.children.nodes.length === 0`. The dispatcher only sees the filtered list and emits no skip event. Meanwhile `crew doctor` uses `fetchRawLinearIssue`, a separate single-issue query that never asks Linear for `children` — so it can't detect the same filter trip. Two queries, two different answers, no visibility for the operator.

## Changes

- **doctor** (`src/commands/ticketDoctor.ts`): new ``Has no sub-issues`` resolution check. When it fails, the verdict flips to `ineligible: parent ticket with sub-issues`. Placed before the repo-cloned check so the structural ticket problem reports first instead of an irrelevant local-env error.
- **boardSource** (`src/lib/boardSource.ts`):
  - `BoardState` gains `parentSkips: ParentSkip[]`; `fetchBoard` populates it from nodes that have children **and** are in their own project's Todo status (Done / In Progress parents aren't a surprising drop and would be noise).
  - `RawLinearIssue` gains `hasChildren: boolean`; `fetchRawLinearIssue`'s GraphQL query now includes `children { nodes { id } }` so doctor can see what `fetchBoard` sees.
- **dispatcher** (`src/commands/dispatcher.ts`): `runOnce` iterates `state.parentSkips` at the start of each tick and emits both a human log line and the structured event `event=dispatch outcome=skipped reason=parent_with_children ticket=<id> children=<n>`, matching the existing skip-event format.

## Why

When a Todo ticket isn't being picked up, the operator's first move is `crew doctor`. Today that diagnostic can confidently report the ticket is eligible while the daemon silently skips it forever — there's no signal anywhere in the system that the ticket is a parent. After this change:

- doctor catches the parent case and reports it as `ineligible`.
- the daemon log emits one `parent_with_children` skip per parent per tick, so tailing the log shows exactly which tickets were passed over and why.

## Test plan

- [x] Red-green TDD for each production change (doctor check, dispatcher logging, fetchBoard parentSkips population).
- [x] `node --run verify` — all 858 tests across 26 files pass, 100% coverage.
- [x] Existing "filters out parent issues that have children" tests in `boardSource.test.ts` and `orchestrator.test.ts` remain green — no behavior change to the issues list, only an addition (`parentSkips`).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@clipboard-health/groundcrew",
-  "version": "3.1.0",
+  "version": "3.1.4",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@clipboard-health/groundcrew",
-      "version": "3.1.0",
+      "version": "3.1.4",
       "license": "MIT",
       "dependencies": {
         "@clipboard-health/clearance": "1.0.8",
--- a/src/commands/cleaner.test.ts
+++ b/src/commands/cleaner.test.ts
@@ -84,7 +84,7 @@ function todoIssue(id: string, overrides: Partial<Issue> = {}): Issue {
 }
 
 function boardOf(issues: Issue[]): BoardState {
-  return { timestamp: "2025-01-01T00:00:00.000Z", issues };
+  return { timestamp: "2025-01-01T00:00:00.000Z", issues, parentSkips: [] };
 }
 
 function hostEntryFor(repository: string, ticket: string): WorktreeEntry {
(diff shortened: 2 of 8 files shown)


therockstorm:
@coderabbitai review

therockstorm:
@mendral-app review

PR #82, opened 2026-05-22
chore: combine watch-loop idle logs and allow app.terraform.io

therockstorm:
## Summary

On every idle watch tick groundcrew was printing two log lines — the dispatcher's idle reason (e.g. `No Todo tickets to pick up`) and then the orchestrator's `Next poll in 120s...` heartbeat on the same timestamp. Fold the heartbeat into the dispatcher's idle messages via an optional `idleSuffix` so each idle tick prints one combined line like `No Todo tickets to pick up; next poll in 120s`. The standalone `Next poll` log is dropped; non-watch (single-shot) runs still print the idle reason with no suffix. Also adds `app.terraform.io` to `clearance-allow-hosts` under Developer tooling.

## Validation

- `node --run verify` — 852 tests pass, 100% coverage.

Agent session: claude --resume 88407ea2-a7d7-4a5f-a421-743c8f402a60

### DIFF
--- a/clearance-allow-hosts
+++ b/clearance-allow-hosts
@@ -65,6 +65,7 @@ registry.npmjs.org
 www.npmjs.com
 
 # Developer tooling
+app.terraform.io
 buf.build
 cdn.playwright.dev
 formulae.brew.sh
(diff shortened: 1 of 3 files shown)


therockstorm:
### Agree

- **src/commands/dispatcher.ts** — Append `idleSuffix` to the remaining idle-return path. Addressed in https://github.com/ClipboardHealth/groundcrew/commit/11c52d5336224b89ebf8593ea3582974ee82747b. The `No eligible Todo tickets after eligibility filtering` log now appends `idleSuffix` like the other three idle branches, so watch mode keeps its `next poll in …s` heartbeat for this case.

```
cae95f8d147a3968
```

PR #87, opened 2026-05-22
feat: per-model preLaunch hook for out-of-sandbox env setup

david-at-clipboard:
## Summary

Adds an optional `preLaunch` hook on `ModelDefinition` — a shell snippet spliced into the launch command **outside** Safehouse/sdx so it runs in the host launch shell with the agent's bootstrap env still in place. Plus a companion `preLaunchEnv: string[]` field that lists env var names to forward into the agent under the default safehouse runner — groundcrew composes them onto the **agent** Safehouse wrap's `--env-pass=` flag, so adopters get `preLaunch` exports through to the agent **and** keep the project's `clearance-allow-hosts` egress allowlist without rewriting `cmd`.

Use case: per-session credential minting (engineer CLI session → scoped machine-identity token → injected env vars) without putting credential-fetch logic inside the sandboxed agent command, where the sandbox would block access to the engineer's credential store (Keychain, file vault, etc.).

```typescript
models: {
  definitions: {
    claude: {
      preLaunch: "SESSION_TOKEN=$(your-mint-command) && export SESSION_TOKEN",
      preLaunchEnv: ["SESSION_TOKEN"],
    },
  },
},
```

The API and shipped docs stay vendor-neutral; `crew.config.example.ts` and the README use a generic `SESSION_TOKEN` export pattern.

## Runner interactions

- **`safehouse` (default)**: `preLaunch` runs on the host immediately after `cd`, **before** any build secrets are sourced into the launch shell. `preLaunchEnv` names are forwarded onto the **agent** Safehouse wrap's `--env-pass=` flag (the setup wrap stays profile-neutral and only forwards build secrets — see "2026-05-28 main merge" below). If `cmd` already starts with `safehouse`, groundcrew can't compose the flag (user owns the wrap) and a **non-empty** `preLaunchEnv` is rejected at config-resolution time with guidance pointing at the manual `--env-pass=` form (an empty `[]` is accepted as a no-op).
- **`none`**: `secrets.env` is sourced, `.groundcrew/setup.sh` runs on the host, build secrets are `unset`, **then** `preLaunch` runs against a clean env, then the agent is exec'd. Same "preLaunch never sees build secrets" contract as safehouse, enforced via post-cleanup ordering instead of pre-source ordering. `preLaunchEnv` is a no-op (exports already inherit).
- **`sdx`**: not supported in v1 — see Limitations.

## Backward compatibility

This is a purely additive change. Users not adopting `preLaunch` / `preLaunchEnv` are unaffected:

- `ModelDefinition.preLaunch` and `ModelDefinition.preLaunchEnv` are both optional — every existing config compiles and validates unchanged.
- A determinism regression test (`is deterministic when preLaunch is undefined`) asserts the launch shell string is stable across calls when both fields are absent.
- `mergeDefinitions` and `validate` only inspect the fields when set — no new required fields, no new validation that can fail on an existing config.
- `failIfLegacyModelKeys` treats both fields the same way as the existing per-model optional fields (`cmd`, `color`, `usage`, `sandbox`) for the `disabled: true` guard.
- No new runtime dependencies, no new binaries on PATH, no new env vars consumed, no new files read at startup. `crew doctor` is unchanged for users not adopting these fields.
- The runner guards (in `prepareAgentLaunch` and `buildLaunchCommand`) only fire when **both** the resolved runner matches **and** the respective field is set. Existing users without `preLaunch` / `preLaunchEnv` see identical behavior.

## Limitations (v1)

- `preLaunch` and a **non-empty** `preLaunchEnv` are not supported when `local.runner` resolves to `sdx`. The sdx runner doesn't forward arbitrary host env vars into the sandbox, so the export contract would silently break — fail fast instead. Validated both in `prepareAgentLaunch` (before any worktree side effects) and in `buildLaunchCommand` (defense in depth). An empty `preLaunchEnv: []` is treated as a uniform no-op in every runner — it forwards zero names, so the unsupported-runner guards do not fire (centralized via `hasPreLaunchEnv` in `src/lib/config.ts` so all four guard sites stay in lockstep).
- A **non-empty** `preLaunchEnv` is rejected when `cmd` already starts with `safehouse` (the user is overriding the default wrap, so groundcrew has no `--env-pass=` flag to inject into). Adopters can fall back to inlining `--env-pass=NAMES` in their own `cmd` — documented in the README as a collapsed fallback section. As of `4e63536`, this is rejected at `prepareAgentLaunch` time (mirror of the existing sdx early-fail) so the operator sees it before the workspace is created. An empty `[]` passes through as a no-op.
- `preLaunchEnv` cannot list any `BUILD_SECRET_NAMES` entry (`NPM_TOKEN`, `BUF_TOKEN`). Those are `unset` on the host between the setup wrap and the agent wrap, so forwarding them via `--env-pass` would silently no-op — rejected at config-load time as of `ca7f13b`.
- Failures in `preLaunch` abort launch — unlike the deps setup step, which logs and continues. Intentional for credential minting: the agent should never start without its bootstrap env in place. Documented as a Gotcha in the README.

## Follow-ups (deferred to future PRs)

- sdx env forwarding via `-e KEY` so `preLaunch` / `preLaunchEnv` work under the sdx runner.
- `crew doctor` warning when `preLaunch` or `preLaunchEnv` is defined but the resolved runner is `sdx`.
- `{{sandbox}}` substitution inside `preLaunch` (currently only `{{worktree}}`).

## Updates since first review pass (2026-05-27)

- **`76ea133` — trap-before-cd + safehouse-path regression test.** Reordered `hostLaunchPrologue` so `trap '…' EXIT` is the first link in the host `&&` chain (matching the sdx chain). Defense in depth against signals between shell start and trap arming, and against future refactors of `setupWithStatusReporting`. Added a `runner: "safehouse"` spawn-based regression mirroring the existing `runner: "none"` one.
- **`c643594` — README clearance trade-off callout** *(superseded by `13aa8e7` — the trade-off it described no longer exists in the recommended path)*.
- **`13aa8e7` — `preLaunchEnv: string[]` field.** Directly addresses the Safehouse env-inheritance review thread — the recommended adoption pattern no longer forces users to override `cmd` with a bare `safehouse` invocation. Groundcrew composes the `--env-pass=` flag onto its own `safehouse-clearance` wrap, so adopters keep both `preLaunch` env forwarding and the project's egress allowlist. The manual `--env-pass=…`-in-`cmd` form remains documented as a collapsed fallback for users who need to override the wrap.
- **`bae8648` — host snippet runs before sourcing build secrets under safehouse.** Reordered the safehouse host chain to `trap → cd → preLaunch → source secrets → read prompt → exec wrap` so the credential-minting snippet never sees `NPM_TOKEN` / `BUF_TOKEN` in env. Directly addresses [@therockstorm's most recent thread](https://github.com/clipboardhealth/groundcrew/pull/87#discussion_r2554260000) — the host chain ordering he flagged is now exactly what the README documents.

## Updates since 2026-05-28 main merge (`aba5027`)

Main shipped two structural changes that this branch had to splice into:

- **#128** ("agent profile shim") — Safehouse now runs setup and the agent in **two separate wraps**. The setup wrap is profile-neutral (`safehouse-clearance ... sh -c '<setup>'`); the agent wrap uses an `mktemp`-d basename symlink to `/bin/sh` (e.g. `…/claude`) so Safehouse selects only the matching agent profile. Build secrets are `unset` on the host between the wraps so they cannot reach the agent.
- **#129** — `sh -lc` → `sh -c` (login-shell init dropped from the wrap exec).

Resolution preserves both contracts simultaneously:

| Phase | What runs | `--env-pass` |
|-------|-----------|--------------|
| Host (pre-wrap) | `trap → cd → preLaunch → source secrets.env → read prompt → rm promptDir` | n/a |
| Setup wrap | `.groundcrew/setup.sh --deps-only` (profile-neutral) | build secrets only |
| Host (between wraps) | `unset NPM_TOKEN BUF_TOKEN; mktemp shim; chained EXIT trap; ln -s /bin/sh shim` | n/a |
| Agent wrap | `exec <agent> "$@"` via the agent-named shim | `preLaunchEnv` only |

**⚠️ Behavior change vs the pre-merge branch:** `preLaunchEnv` names are now forwarded **only to the agent wrap**, not to the setup wrap. `.groundcrew/setup.sh` no longer sees `preLaunchEnv` vars under safehouse. Strictly better isolation — aligned with #128's least-privilege intent that the setup phase doesn't carry agent-side grants. If your `setup.sh` legitimately needs one of these vars, source it in `setup.sh` itself or use `runner: "none"`.

The combined EXIT trap (`rm shim_dir; rm promptDir`) covers both targets so there's never an unarmed window between shim creation and the post-wrap cleanup.

**`502baca6` follow-up:** grouped the final Safehouse agent-wrap status/cleanup block so it only runs after the `&&` chain reaches the agent wrap. This keeps the EXIT trap armed when `preLaunch` returns non-zero without calling `exit`, matching the documented fail-closed credential minting shape.

**`37e1066` follow-up:** Scrubbed the listed `preLaunchEnv` names in the same pre-hook cleanup as build-secret names, so stale same-named ambient credentials are not forwarded when the hook does not overwrite them.

## Test plan

- [x] `node --run verify` green — 1024 tests, 100% statements / branches / functions / lines.
- [x] Determinism: `preLaunch` undefined → launch script is identical across calls (`is deterministic when preLaunch is undefined`).
- [x] Safehouse ordering: `trap → cd → preLaunch → source secrets → read prompt → setup wrap → host unset → shim → agent wrap`. preLaunch precedes any build secrets in env.
- [x] Two-wrap isolation: setup wrap forwards build secrets only; agent wrap forwards `preLaunchEnv` only. The old single-wrap composition (`NPM_TOKEN,BUF_TOKEN,SESSION_TOKEN`) does not appear anywhere in the launch string.
- [x] EXIT-trap cleanup: `preLaunch: "exit 7"` (runner `none`) and `preLaunch: "exit 9"` (runner `safehouse`) abort launch and wipe `promptDir` via the trap. Combined trap (`rm shim_dir; rm promptDir`) replaces the shim-only trap so a failure between shim creation and the agent wrap still wipes both. A Safehouse `preLaunch` that returns non-zero without calling `exit` (`SESSION_TOKEN=20 20 12 61 79 80 81 98 100 701 33 204 250 395 398 399 400false) && export SESSION_TOKEN`) also leaves the trap armed and wipes `promptDir`.
- [x] Early guards (`prepareAgentLaunch`): throws before worktree creation when `sdx` + `preLaunch`, `sdx` + non-empty `preLaunchEnv`, or `safehouse` + non-empty `preLaunchEnv` + safehouse-prefixed cmd. Empty `preLaunchEnv: []` passes through as a no-op in every case (locked by `sdx` + `[]` test in `setupWorkspace.test.ts`).
- [x] Late guards (`buildLaunchCommand`): throws when `sdx` + `preLaunch`, `sdx` + non-empty `preLaunchEnv`, or `safehouse-prefixed cmd` + non-empty `preLaunchEnv` (defense in depth). Empty `preLaunchEnv: []` is a no-op on both paths (locked by sdx and safehouse-prefixed-cmd `[]` tests in `launchCommand.test.ts`).
- [x] `--env-pass=` composition (per wrap): build-secrets-only on setup wrap, `preLaunchEnv`-only on agent wrap, both wraps unflagged when neither applies.
- [x] Config validation: `preLaunch` and `preLaunchEnv` merge through config overlays; non-array `preLaunchEnv` rejected; invalid env-var name entries rejected with index; entries overlapping `BUILD_SECRET_NAMES` rejected with guidance; empty array no-op; combining with `disabled: true` rejected.
- [x] End-to-end smoke against a local secrets backend: ephemeral identity created, minted JWT carries the expected identity, secrets injected into the agent, no `.env` on disk, identity deleted on EXIT.
- [x] Live `crew start <ticket>` with a `preLaunch` snippet + `preLaunchEnv: ["SESSION_TOKEN"]` (default `cmd: "claude"`): minted env var visible inside the Safehouse sandbox via groundcrew-composed `--env-pass=`; no bootstrap creds leak through; cleanup fires on workspace close.

Agent session: `codex resume 019e705a-3700-78c1-a36a-a2770e7a195a`

Agent session: `codex resume 019e7062-aff0-7721-b487-b5a29e9f9842`

<sub>🤖 <code>commit-push-pr:created v1 core@3.5.0</code></sub>


README.md
```diff
@@ -125,6 +125,23 @@ Set them in the shell you run `crew` from. Anything not in this list is ignored
...
+Build secrets shuttle build-time values _into_ setup. `preLaunch` does the opposite for the _agent_ phase: it runs a host-shell snippet **outside** Safehouse/sdx, after build-secret cleanup and before the prompt is read, so exports are inherited by the wrapped `cmd`. Use this when the agent needs a short-lived credential that has to be minted from something the sandbox can't reach (e.g. an engineer CLI session in Keychain), and you don't want any of that source material in the agent's process.
+
+```typescript
+models: {
+  definitions: {
+    claude: {
+      preLaunch: [
+        "SESSION_TOKEN=$(your-mint-command)",
+        "export SESSION_TOKEN",
+      ].join("; "),
```
therockstorm: This example does not fail closed as written. In `sh`, `SESSION_TOKEN=$(your-mint-command); export SESSION_TOKEN` returns the status of `export`, so a failed mint can still make `preLaunch` return 0 and the agent starts without the token. I verified the shape locally with `sh -c 'TOKEN=$(false); export TOKEN'`, which exits 0.

For a credential-minting example, please show a fail-closed form such as `SESSION_TOKEN=$(your-mint-command) && export SESSION_TOKEN` or `SESSION_TOKEN=$(your-mint-command) || exit $?; export SESSION_TOKEN`.

src/lib/launchCommand.ts
```diff
@@ -136,6 +145,9 @@ function buildHostLaunchCommand(arguments_: LaunchCommandArguments): string {
   if (arguments_.secretsFile !== undefined) {
     lines.push(unsetSecretsLine());
   }
+  if (arguments_.definition.preLaunch !== undefined) {
+    lines.push(renderPreLaunch(arguments_.definition.preLaunch, arguments_.worktreeDir));
```
therockstorm, line 393: If `preLaunch` exits nonzero here, the generated `&&` chain stops before `_p=$(cat ...)` and `rm -rf ${promptDir}` run. That leaves the staged prompt directory behind, including `prompt.txt` and any staged `secrets.env`, when the credential bootstrap fails inside the workspace shell.

Please put prompt/secrets cleanup behind an `EXIT` trap or otherwise make it run on the failure path before `preLaunch` can abort the launch script.

README.md
```diff
@@ -125,6 +125,23 @@ Set them in the shell you run `crew` from. Anything not in this list is ignored
 3. After setup completes, the script `unset`s every name in `BUILD_SECRET_NAMES` and then `rm -rf`s the entire prompt dir (including `secrets.env`) before `exec`'ing the agent. See `sourceSecretsLine` / `unsetSecretsLine` and the `rm -rf` / `exec` lines in [`src/lib/launchCommand.ts`](./src/lib/launchCommand.ts). The rollback path on setup failure ([`src/commands/setupWorkspace.ts`](./src/commands/setupWorkspace.ts)) wipes the prompt dir too.
 
 Net effect: by the time the agent process exists, the values are gone from the environment and the file is gone from disk.
+Build secrets shuttle build-time values _into_ setup. `preLaunch` does the opposite for the _agent_ phase: it runs a host-shell snippet **outside** Safehouse/sdx, after build-secret cleanup and before the prompt is read, so exports are inherited by the wrapped `cmd`. Use this when the agent needs a short-lived credential that has to be minted from something the sandbox can't reach (e.g. an engineer CLI session in Keychain), and you don't want any of that source material in the agent's process.
```
therockstorm: This is too broad for the default Safehouse path. Safehouse documents a sanitized runtime environment by default, with env behavior controlled explicitly via full `--env`, named `--env-pass=...`, or file-based `--env=./agent.env` modes. So exports from `preLaunch` inherit into the host launch shell, but they will not necessarily reach the agent when the default Safehouse wrapper is used.

Please make the Safehouse requirement first-class in the README/example, e.g. show `safehouse ... --env-pass=SESSION_TOKEN ...` or state directly that the generic `preLaunch` export example only reaches the agent under `runner: "none"` unless the Safehouse command forwards the variable.

Sources checked: https://agent-safehouse.dev/docs/overview and https://www.agent-safehouse.dev/docs/usage#environment-modes

therockstorm (commented):
I verified the Safehouse env behavior against the upstream docs and checked the shell failure modes locally. Three comments.

therockstorm (changes requested):
See comments.

README.md
david-at-clipboard: FGood catch — fixed in `fa7846f`. The README and example now show the fail-closed form:

```ts
preLaunch: "SESSION_TOKEN=$(your-mint-command) && export SESSION_TOKEN",
```

Prose right under the snippet explains why the `&&` is required:

> "`&&` ensures `export` only runs when the mint succeeded; a failed mint propagates non-zero out of `preLaunch` and aborts launch before the agent starts."

Also de-duped the redundant example in `crew.config.example.ts` per CodeRabbit's adjacent comment in the same pass.

src/lib/launchCommand.ts
david-at-clipboard: Fixed in `fa7846f`: added a `trapCleanupLine` helper that arms `trap 'rm -rf <promptDir>' EXIT` early in every chain, so the failure path is covered without depending on the success-path `rm -rf` running.

Strengthened further in `76ea133` after another look:

- The trap is now armed **before** `cd` (matching the sdx chain), so a signal or a future refactor of `setupWithStatusReporting` can't open a window where the trap isn't yet active.
- Added a `runner: "safehouse"` spawn-based regression (`preLaunch: "exit 9"` → expect status 9 + `promptDir` is `ENOENT`) alongside the existing `runner: "none"` one in `src/lib/launchCommand.test.ts` - so both host paths have end-to-end coverage of the trap firing on `preLaunch` failure.

README.md
david-at-clipboard: Fixed in `13aa8e7`: added `preLaunchEnv: string[]` on `ModelDefinition`. Groundcrew appends the names to its own `safehouse-clearance` wrap's `--env-pass=` flag (composing with `BUILD_SECRET_NAMES` when present), so adopters keep both `preLaunch` forwarding and the default clearance allowlist without rewriting `cmd`:

```typescript
claude: {
  preLaunch: "SESSION_TOKEN=$(your-mint-command) && export SESSION_TOKEN",
  preLaunchEnv: ["SESSION_TOKEN"],
},
```

README rewritten around this pattern; the manual `safehouse --env-pass=` form is kept as a collapsed fallback for `cmd`-owns-the-wrap cases. Same sdx guards as `preLaunch`, plus a launch-time reject when `cmd` already starts with `safehouse`.

david-at-clipboard:
> See comments.

@therockstorm - thanks for the review - I went through and resolved them - a number of changes (positive ones) needed to occur after #124, so I did another review.

Let me know if we're good to merge. :)

therockstorm:
@david-at-clipboard I think there is a separate ordering issue with build secrets for Safehouse. `hostLaunchPrologue` sources `secrets.env` into the host shell, and then `preLaunchPromptAndExec` runs `preLaunch` before the `safehouse-clearance` exec. The build secrets are only unset inside the wrapped shell after setup, so `preLaunch` can still see `NPM_TOKEN`/`BUF_TOKEN` when a `secretsFile` is staged.

That contradicts the documented contract that `preLaunch` runs after build-secret cleanup, and it broadens secret exposure to the host-side credential minting snippet.

Could we split the Safehouse path so it runs roughly:

1. trap + `cd`
2. `preLaunch`
3. source `secrets.env`
4. read prompt / `rm -rf`
5. `exec safehouse-clearance --env-pass=BUILD_SECRETS,preLaunchEnv ...`

That keeps `preLaunchEnv` working while preventing `preLaunch` from seeing build-time setup secrets.

src/lib/agentLaunch.ts
```diff
@@ -41,6 +41,22 @@ export async function prepareAgentLaunch(input: {
...
+    throw new Error(
+      `Local groundcrew ${input.purpose} with the sdx runner do not support preLaunch on model '${input.model}'. ` +
+        "Use local.runner 'safehouse' or 'none', or remove preLaunch from the model.",
+    );
+  }
+  if (
+    runner === "sdx" &&
+    input.definition.preLaunchEnv !== undefined &&
+    input.definition.preLaunchEnv.length > 0
+  ) {
```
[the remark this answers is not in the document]
david-at-clipboard: Declining — keeping the `length > 0` gate intentional, and centralizing it so this stays consistent.

`preLaunchEnv: []` forwards zero names, so it's a no-op in every runner. Rejecting `[]` only on `sdx` would make an empty list *accepted* under `safehouse`/`none` but *fatal* under `sdx` / safehouse-prefixed `cmd` — that's a worse asymmetry than the one flagged. The parity-with-`preLaunch` framing doesn't hold either: `preLaunch` is a string with no empty-no-op form (whitespace-only is already rejected by the non-whitespace validator).

Follow-up applied:

- Centralized the predicate into a single exported `hasPreLaunchEnv(definition)` helper in `src/lib/config.ts` so the four guard sites (`agentLaunch.ts` sdx + safehouse-prefixed-cmd; `launchCommand.ts` sdx + safehouse-prefixed-cmd) stay in lockstep mechanically — no more per-site duplication to drift.
- Added the `sdx` + `preLaunchEnv: []` no-op test in `src/commands/setupWorkspace.test.ts` you asked for, plus matching `sdx` + `[]` and safehouse-prefixed-cmd + `[]` no-op tests in `src/lib/launchCommand.test.ts`, so the empty-array contract is locked at both the prepare boundary and the launch-command boundary.
- Tightened README + PR description to say only a **non-empty** `preLaunchEnv` is rejected; an empty `[]` is a uniform no-op in every runner.

therockstorm:
@coderabbitai review

therockstorm:
@mendral-app review

david-at-clipboard:
@coderabbitai review

david-at-clipboard:
@mendral-app review

therockstorm (approved):

david-at-clipboard:
@therockstorm - appreciate your review and additions - I'll get this merged and keep pushing on the actual Infisical parts.

PR #78, opened 2026-05-22
[HRD-452] Add pluggable ticket-source foundation

paulbaranowski:
## Summary

Adds a pluggable ticket-source foundation so groundcrew can construct and verify multiple ticket sources from `crew.config.ts` via `sources: SourceConfig[]`. The PR introduces `TicketSource`, `Board`, a runtime-scanned adapter registry, source config validation, a Linear adapter over the existing Linear machinery, and a shell adapter for JSON-speaking external commands.

Linear dispatch behavior is intentionally unchanged: the current tick loop still reads through `boardSource.ts`. This PR verifies configured sources at orchestrator startup and leaves the consumer-loop migration to `board.fetch()` for a focused follow-up.

The shell adapter now has bounded stdout/stderr buffering, timeout cleanup that terminates the process group on non-Windows, Zod-validated JSON I/O, positive-integer timeout validation, command placeholders, and README coverage for the command contract.

## What's in scope

- `TicketSource` interface - `verify`/`fetch`/`resolveOne`/`markInProgress` (`src/lib/ticketSource.ts`)
- `Board` composer - fans calls across N adapters, source-prefixed canonical ids, `AmbiguousTicketError` on natural-id collisions (`src/lib/board.ts`)
- `AdapterDefinition` + runtime directory-scanned registry - drop a directory in `src/lib/adapters/`, the registry picks it up at startup via `readdirSync` + dynamic `import()` (`src/lib/adapterDefinition.ts`, `src/lib/adapters/registry.ts`)
- `buildSources` - config to `TicketSource[]` dispatch via the registry, threading an `AdapterContext` with global `ResolvedConfig` into each adapter's `create()` (`src/lib/buildSources.ts`)
- Linear adapter - wraps existing `boardSource.ts` machinery; converts Linear-shaped issues to canonical issues with per-project status mapping (`src/lib/adapters/linear/`)
- Shell adapter - subprocess execution with `${id}`/`${canonicalId}`/`${name}` shell-quoted substitution, configurable positive-integer timeouts, Zod-validated JSON I/O, optional commands with fallbacks, bounded output, and process-group timeout cleanup (`src/lib/adapters/shell/`)
- Config - `sources: SourceConfig[]` on `Config` + `ResolvedConfig`; structural validation in `config.ts`, per-adapter Zod validation in `buildSources` (`src/lib/config.ts`)
- Orchestrator - calls `buildSources(config.sources, {globalConfig})` + `board.verify()` at startup so configured sources have their preflight exercised before the first tick
[...]

PR #89, opened 2026-05-23
feat(tickets): pluggable ticket system MVP-2 — consumer refactor

paulbaranowski:
## Summary

Picks up where PR #78 stopped. PR #78 shipped the **pluggable ticket-source abstraction** (canonical `Issue` shape, `TicketSource` interface, `Board` composer, Linear adapter, shell adapter, runtime registry, config-driven dispatch) but consumer code (`dispatcher`, `cleaner`, `eligibility`, `setupWorkspace`, `ticketDoctor`, `doctor`, `orchestrator`) was still wired to the legacy `boardSource.ts` and branched on raw Linear status strings. This PR completes the migration:

- Every consumer reads the canonical `Issue`/`Blocker` shape
- Writeback goes through `board.markInProgress(issue)`, routed by source
- Repository validation moves to a host-level helper that WARN+skips unknown repos uniformly across sources
- The orchestrator's tick path reads exclusively from `board.fetch()`
- `boardSource.ts` and `linearIssueStatus.ts` are deleted; their internals split into focused per-file modules under `src/lib/adapters/linear/`

After this PR, shell-source tickets actually flow through dispatch: configure `sources: [{ kind: "shell", ... }]` and the dispatcher treats those tickets the same as Linear tickets.

> **Rebased onto main's PR #110.** That PR dropped the entire `config.linear` block and switched Linear classification to workflow `state.type` (viewer + `agent-*` label filter at the GraphQL layer). This PR integrates onto that new model: the pluggable architecture (canonical Issue, Board, adapter pattern) survives intact; the Linear adapter's internals were rewritten on `state.type` classification rather than per-project status names. `statusMap.ts` was deleted (classification is now an inline switch in `factory.ts`).

## Locked design decisions

| # | Decision | Choice |
|---|---|---|
| 1 | Bridge between legacy/canonical Issue, or full migration? | **Canonical migration.** No bridge code; every consumer reads canonical end-to-end. TypeScript catches every mismatch at compile time. |
| 2 | Where to synthesize the implicit Linear source? | **Helper at runtime** (`sourcesFromConfig`). Post-#110 Linear is always-on, so the helper synthesizes a Linear source by default and omits it only when the user already declared one with `kind: linear` or `name: "linear"`. `ResolvedConfig.sources` stays exactly what the user typed. |
| 3 | What happens to an issue whose `repository` isn't in `workspace.knownRepositories`? | **WARN-log + skip** at the dispatcher, uniformly across sources. **Behavior change**: the legacy `RepositoryResolutionError`-on-unknown-repo throw is removed. |
| 4 | How does `ticketDoctor` get Linear-specific behavior in a source-agnostic abstraction? | **Optional `TicketSource` methods** (`refreshBlockers?`, `countInProgress?`) + Board fan-out wrappers with built-in fallbacks. Plus `Board.sources()` for display iteration. |
| 5 | `crew setup migrate` CLI for legacy → unified config rewrite? | **Out of scope.** Post-#110 `linear.projects` no longer exists, so no legacy migration path is needed. |

## What landed

| Scope | Touches |
|---|---|
| Foundations | `sourcesFromConfig` helper, `Board.sources()` getter, `description` on board snapshot, canonical fixture helpers, `LinearSourceRef` exported |
| setupWorkspace | Drops `fetchTicket` (no more Linear re-fetch from dispatch path); `setupWorkspaceCli` routes through `board.resolveOne`; `details` is now required |
| cleaner | Migrates to canonical; orchestrator computes canonical state alongside legacy via `toCanonicalIssue`; extracts `naturalIdFromCanonical` helper |
| eligibility | Takes canonical `Blocker`; branches on `status === "done"`; `crew cleanup` hint uses natural id |
| TicketSource extensions | Optional `refreshBlockers?` / `countInProgress?` on TicketSource; Board fan-out wrappers with fallbacks; Linear adapter implements both |
| dispatcher | `DispatcherDeps` takes `Board` (not `LinearClient`); canonical status checks; `board.markInProgress` writeback; `dispatchableRepository` helper; `parseRepository` returns asserted name on unknown-repo |
| ticketDoctor | Per-ticket fetch + decision logic on Board API; preserves `rawHasMoreBlockers` safety gate; `isGroundcrewIssue` guard replaces unsafe cast; deletes `resolveRepositoryFromDescription` duplicate and calls `resolveRepositoryFor` directly |
| doctor + orchestrator finalization | `checkLinearApiKey` → `checkSourceProbe` (via `board.verify()`); `--no-source-probe` flag with `--no-linear` alias; orchestrator drops legacy `boardSource` and dual-state |
| Reorganize + delete | Move boardSource internals into `adapters/linear/{fetch,parsing,writeback,client}.ts`; move `getLinearClient` from `util.ts`; delete `boardSource.ts` + `linearIssueStatus.ts`; split tests |

## End-state file structure

```
src/lib/adapters/linear/
  factory.ts        # Assembles the Linear TicketSource from siblings.
                    # Hosts canonicalStatusFromStateType (state.type → canonical)
  index.ts          # Public surface (LinearSourceRef, etc.)
  schema.ts         # Adapter config schema (Zod)
  client.ts         # NEW — getLinearClient, lazyLinearClient (was util.ts)
  fetch.ts          # NEW — verifyViewer, fetchBoard, fetchResolvedIssue,
                    #       fetchBlockersForTicket, fetchInProgressIssueCount,
                    #       fetchRawLinearIssue, createBoardSource,
                    #       isIssueTodo/isIssueInProgress/isTerminalStateType/
                    #       isTerminalStatusForIssue/isTerminalStatusForBlocker
  parsing.ts        # NEW — resolveModelFor, resolveRepositoryFor, parseRepository,
                    #       buildRepositoryRegex (shared between dispatch + manual setup)
  writeback.ts      # NEW — createLinearIssueStatusUpdater (state.type === "started" lookup)
src/lib/
  board.ts          # +sources(), +refreshBlockers(), +countInProgress()
  ticketSource.ts   # +refreshBlockers?/countInProgress? on TicketSource;
                    #  naturalIdFromCanonical helper
  buildSources.ts   # +sourcesFromConfig() helper
  repositoryValidation.ts  # NEW — dispatchableRepository(issue, knownRepos, log)
  testing/canonicalFixtures.ts  # NEW — canonicalLinearIssue, canonicalBlocker
src/testHelpers/
  boardFixtures.ts  # NEW — makeBoard() shared test factory
(deleted: src/lib/boardSource.ts, boardSource.test.ts, linearIssueStatus.ts, linearIssueStatus.test.ts)
```

## Behavior changes worth flagging

1. **Unknown-repo dispatch is now WARN+skip, not throw.** Today: an `agent-typo-repo` label whose repo isn't in `workspace.knownRepositories` raises `RepositoryResolutionError` and crashes the tick. After this PR: the dispatcher logs `issue linear:eng-123 references unknown repository typo/repo; configured workspace.knownRepositories: …` and skips the ticket for the tick. Loud enough to debug, quiet enough not to crash a multi-source tick.

2. **`Blocker.status` is no longer nullable.** Legacy `Blocker.status: string | undefined` had two failure modes — `undefined` (Linear didn't return state for the blocker) and a `state.type` the adapter can't classify (e.g. `backlog`, `triage`). Both were treated identically as "still blocking" by `isTerminalStatusForBlocker`. Canonical `Blocker.status: CanonicalStatus` collapses both into `"other"` at the adapter boundary with `statusReason: "missing"` vs `"unmapped"` for diagnostics, so consumers don't have to handle two failure modes separately. Same effective semantics; the type is now honest about what consumers actually do with it. The raw native value is preserved on `Blocker.nativeStatus` when the source provided one.

3. **`--no-linear` is renamed to `--no-source-probe` in `crew doctor` host mode.** `--no-linear` still works as an alias.

4. **Issue ids are source-prefixed end-to-end.** `linear:eng-220` everywhere internally; the dispatcher strips the prefix via `naturalIdFromCanonical()` at the boundary where worktree-naming or setupWorkspace's natural-id contract needs it.

5. **`crew doctor --no-fetch` is now rejected in host mode.** Previously, `--no-fetch` was accepted into the argument list but only consumed by ticket-mode (`--ticket <id>`) invocations, so `crew doctor --no-fetch` without a ticket was a silent no-op. The CLI now throws `crew doctor: --no-fetch requires --ticket` so the flag's intent can't be silently dropped.

6. **Classification is `state.type`-driven, not status-name-driven** (inherited from main's PR #110, re-applied through the pluggable layer). The canonical mapping is `unstarted → todo`, `started → in-progress`, `completed | canceled | duplicate → done`, anything else → `other`. Workspaces that renamed status columns ("Todo" → "To Do", "Done" → "Shipped") Just Work without per-team config; canceled and duplicate states are correctly treated as terminal.

## Test strategy

Every consumer test file migrates from legacy `Issue` literals (with `uuid`, `statusId`, `teamId`, raw status strings) to `canonicalLinearIssue(...)` fixtures. The `as LinearSourceRef` cast is the documented escape hatch — used in two places (Linear adapter writeback + ticketDoctor's raw-payload rendering). Both call sites have `oxlint-disable` justifications referencing the adapter contract.

The boundary check (canonical `Issue` carries `description` populated from the board snapshot) is enforced in `factory.test.ts` and prevents the dispatch path from ever re-fetching tickets to fill `description`.

## Verification

- **1218 tests pass**, **100% coverage** maintained throughout (statements / branches / functions / lines)
- `node --run verify` passes all 10 checks (architecture, build, cpd, format, knip, lint, markdown, spell, syncpack, test)


paulbaranowski:
## Fixes Applied Successfully

Fixed 2 file(s) based on 2 CodeRabbit feedback item(s).

**Files modified:**
- `src/lib/adapters/linear/factory.ts` — centralized canonical-id construction via `toCanonicalId(sourceName, naturalId)`
- `src/lib/buildSources.ts` — `sourcesFromConfig` now throws when an explicit `sources[]` entry resolves to runtime name `"linear"`, instead of letting the collision surface as a deeper `createBoard` error

Also added regression tests in `src/lib/buildSources.test.ts` for the new collision check (kind-only, explicit name, distinct name, empty-projects).

**Commit:** `3b13123`

The latest autofix changes are on the `emdash/pluggable-ticket-system-v2-8rmm9` branch.

paulbaranowski:
## Fixes Applied Successfully

Fixed 3 file(s) (+1 regression test) based on 3 CodeRabbit feedback item(s).

**Files modified:**
- `src/cli.ts` — reject `crew doctor --no-fetch` in host mode
- `src/cli.test.ts` — regression test for the new validation
- `src/commands/dispatcher.ts` — repo validation runs before slot allocation
- `src/lib/adapters/shell/invoke.ts` — Buffer-accurate byte truncation in shell adapter

**Commit:** `8db66db`

The latest autofix changes are on the `emdash/pluggable-ticket-system-v2-8rmm9` branch.

paulbaranowski:
## Fixes Applied Successfully

Fixed 1 file (+1 regression test) based on 1 CodeRabbit feedback item.

**Files modified:**
- `src/commands/dispatcher.ts` — move repository validation ahead of usage() HTTP call and workspaces.probe shell-out
- `src/commands/dispatcher.test.ts` — regression test pinning the new ordering

**Commit:** `ad0c29f`

The latest autofix changes are on the `emdash/pluggable-ticket-system-v2-8rmm9` branch.

src/lib/adapters/linear/parsing.ts
```diff
@@ -0,0 +1,171 @@
...
+}): RepositoryResolution {
+  const { description, config } = arguments_;
+  if (description === undefined || description.length === 0) {
+    return { kind: "missing" };
+  }
+  const repository = buildRepositoryRegex(config).exec(description)?.[1];
+  if (repository === undefined) {
+    return { kind: "missing" };
+  }
+  return { kind: "ok", repository };
```
therockstorm: This regresses canonical repository resolution for manual setup. The previous `resolveRepositoryFor` resolved a bare match like `repo` back to the exact `workspace.knownRepositories` entry, e.g. `org/repo`, and treated ambiguous bare names as missing. Now `fetchResolvedIssue()` can return `repository: "repo"`, so `crew run --ticket ...` launches against the wrong repo path when `knownRepositories` contains `org/repo`. Please share the same canonicalization and ambiguity handling as `parseRepository`, with tests for bare and ambiguous matches.

src/lib/adapters/linear/factory.ts
```diff
@@ -202,12 +202,27 @@ export function createLinearTicketSource(
     async markInProgress(issue: CanonicalIssue): Promise<void> {
       // oxlint-disable-next-line typescript/no-unsafe-type-assertion -- by the Linear adapter's contract, every Issue it produces carries a LinearSourceRef in sourceRef
       const ref = issue.sourceRef as LinearSourceRef;
-      await issueStatusUpdater.markInProgress({
+      await getIssueStatusUpdater().markInProgress({
```
therockstorm, line 232: The Linear writeback negative cache is no longer reset after the Board migration. Before this PR, `createDispatcher.runOnce()` called `issueStatusUpdater.resetMissingInProgressCache()` each tick; now writeback is routed through `Board` and no caller resets the cached updater. A temporary Linear workflow/config mismatch can therefore poison a watch session until process restart. Please either expose/reset this through a Board/source tick lifecycle hook or remove/change the negative cache.

src/commands/ticketDoctor.ts
```diff
@@ -379,20 +431,65 @@ function buildModelChecks(raw: RawLinearIssue, config: ResolvedConfig): ModelRes
...
+    .map((value) => value.replaceAll(/[.*+?^${}()|[\]\\]/g, String.raw`\$&`))
+    .join("|");
+  const repositoryRegex = new RegExp(String.raw`\b(${alternation})\b`);
+  const matched = repositoryRegex.exec(description)?.[1];
+  if (matched === undefined) {
+    return { kind: "missing" };
+  }
+  // Canonicalize: resolve the matched value (which may be a bare repo name)
+  // back to the full "owner/repo" entry from knownRepositories (PR #77).
+  const canonical = config.workspace.knownRepositories.find(
```
therockstorm: `ticketDoctor` now duplicates Linear repository parsing, and this copy can diverge from the adapter/dispatcher behavior. In particular, `.find()` silently chooses the first `workspace.knownRepositories` entry for ambiguous bare repo names, while `parseRepository` treats that ambiguity as a failure. Please reuse the Linear adapter resolver instead of duplicating the regex/canonicalization here, and add coverage for ambiguous bare repo names so diagnostics match dispatch behavior.

therockstorm (changes requested):
I like the overall direction: canonical Issue/Board boundaries, adapter-owned source details, and dispatcher consumers no longer hard-wired to Linear.

src/lib/adapters/linear/parsing.ts
paulbaranowski: Fixed in 747712d. Extracted a shared `canonicalizeRepositoryMatch` helper used by both `parseRepository` and `resolveRepositoryFor`, so a bare `repo-a` match canonicalizes back to the configured `org/repo-a` entry and ambiguous bare names surface as `missing` (which `fetchResolvedIssue` turns into a `RepositoryResolutionError` — same conflation `parseRepository` already uses). Added two regression tests on `resolveRepositoryFor`: bare-match canonicalization and ambiguous-bare-match.

src/lib/adapters/linear/factory.ts
paulbaranowski: Fixed in 4ec2520. Removed the negative cache entirely (no `beginTick`-style hook). Slot count caps `markInProgress` per tick at 1-5, so re-fetching team workflow states on each failing attempt costs a handful of extra Linear API calls; in return the operator-side fix is picked up without a process restart. Positive cache for successfully-resolved state ids stays. Rewrote the previous negative-cache tests to pin the new re-fetch-on-every-failure behavior.

src/commands/ticketDoctor.ts
paulbaranowski: Fixed in 4ec2520. Deleted the inline `resolveRepositoryFromDescription` and now call the canonical `resolveRepositoryFor` from `parsing.ts` directly, so the doctor and the dispatch path agree on ambiguous bare names (both surface as missing / `RepositoryResolutionError`). Moved the empty-knownRepositories guard down into the shared `canonicalizeRepositoryMatch` helper so every consumer is protected from the `/\b()\b/` empty-match bug. Added a regression test in `ticketDoctor.test.ts` exercising the ambiguous-bare-name case end-to-end through the doctor, and a matching coverage test in `parsing.test.ts` for the empty-repos guard.

paulbaranowski:
## Fixes Applied Successfully

Fixed 1 file (+1 regression-test file) based on 1 CodeRabbit feedback item.

**Files modified:**
- `src/lib/testing/canonicalFixtures.ts` — `canonicalLinearIssue`, `canonicalBlocker`, and `canonicalShellIssue` now all route through `toCanonicalId` so fixture-side canonical ids match the runtime's lowercased form
- `src/lib/testing/canonicalFixtures.test.ts` — added regression tests pinning that uppercase `naturalId` values are lowercased in the returned canonical id

**Commit:** `e383f11`

The latest autofix changes are on the `emdash/pluggable-ticket-system-v2-8rmm9` branch.

paulbaranowski:
@therockstorm could you please review again?

therockstorm:
@paulbaranowski  I'm doing another refactor first and will then get this PR merged.

therockstorm:
Closed in https://github.com/ClipboardHealth/groundcrew/pull/119

PR #91, opened 2026-05-23
feat: add 'crew upgrade' subcommand with availability nudge

paulbaranowski:
## Summary

Adds a self-upgrade command for `crew` so users don't have to remember `npm install -g @clipboard-health/groundcrew@latest`, plus a passive nudge on every other subcommand when a newer version is published.

- **`crew upgrade [<version>] [--check]`** — installs `@latest` by default, accepts an exact version (upgrade or downgrade), or reports status with `--check`. Refuses cleanly when not running from a global npm install (`npm link` dev, npx temp, project dependency, unknown location) — each refusal carries a tailored hint pointing at the right fix.
- **Passive nudge** prints a single line to **stderr** before every non-`upgrade` subcommand when the registry's `latest` exceeds the running version:
  ```
  [crew] 3.2.0 available — run `crew upgrade` (you have 3.1.8)
  ```
  Cached at `~/.cache/groundcrew/upgrade-check.json` with a 6 h TTL and a 1000 ms hard fetch timeout. Explicit `crew upgrade` / `crew upgrade --check` calls also prime the cache, so the next subcommand never pays a network call. Silenced by `GROUNDCREW_NO_UPGRADE_CHECK=1`; honors `npm_config_registry` for corporate mirrors.
- **No new runtime dependencies** — version compare, registry fetch, and cache are hand-rolled (~50 lines each) against the existing `host.ts` / `commandRunner.ts` helpers. No `semver`, no chalk.

## Design decisions

| Decision | Outcome |
|---|---|
| Distribution | npm self-upgrade only (Homebrew considered, rejected) |
| Channel | Stable `latest` dist-tag only (no `--beta` / pre-release) |
| Downgrade | `crew upgrade <ver>` where `<ver>` < current prints `"downgrading A → B"` and proceeds (no prompt, no `--force`) |
| Prereleases | SemVer 11.3 ordering: stable > prerelease of same numeric, so a beta build correctly sees the upgrade nudge. Coarse comparison only (two prereleases of same numeric are equal). |
| Nudge cadence | Every non-`upgrade` invocation (incl. `crew run`), cached |
| Nudge style | Plain text, stderr, single line, npm-style (top of output) |
| Non-global refusal | Hard refuse; each install-kind gets a tailored hint |
| Post-upgrade changelog | Out of scope for v1 |

## Files

- **New** `src/lib/upgrade.ts` (+ test) — `parseVersion`, `compareVersions`, `fetchLatestVersion`, cache helpers, `composeNudgeMessage`, `computeUpgradeNudge`.
- **New** `src/lib/npmGlobal.ts` (+ test) — `classifyInstall`, `runNpmInstallGlobal`, `createDefaultNpmSpawner`, plus `detectInstallPath` / `detectNpmRootGlobal` / `detectIsSymlink` helpers.
- **New** `src/commands/upgrade.ts` (+ test) — `upgradeCli(argv, options)` orchestration and `createDefaultUpgradeCliOptions` factory.
- **Modified** `src/cli.ts` — registers the `upgrade` subcommand and calls the nudge before dispatch.
- **Modified** `src/cli.test.ts` — verifies dispatch routing and nudge wiring.

## Test plan

- [x] `node --run verify` passes — 1094 tests, 100% coverage across statements, branches, functions, lines.
- [ ] Manual smoke against a published version mismatch:
  - [x] `crew upgrade --check` from a globally-installed `crew` on an older version → prints `"X.Y.Z available …"`.
  - [x] `crew upgrade --check` when current = latest → prints `"crew is up to date (X.Y.Z)"`.
  - [x] `crew doctor` from an older version → nudge appears once on stderr, doctor output unaffected.
  - [ ] `GROUNDCREW_NO_UPGRADE_CHECK=1 crew doctor` → no nudge.
  - [ ] `crew upgrade` from the dev checkout (`node --run crew -- upgrade`) → refuses with `"npm link"`/`"project dependency"` hint and exits 1.
  - [ ] `crew upgrade 3.1.5` from a 3.1.8 install → prints `"downgrading 3.1.8 → 3.1.5"` and downgrades.

## Notes for the reviewer

- The `runInstall` closure in `createDefaultUpgradeCliOptions` exists so tests can swap out the real `runNpmInstallGlobal` + `createDefaultNpmSpawner` plumbing without spawning real `npm` processes.
- I considered passing `now: Date.now` directly to `computeUpgradeNudge` vs. wrapping in an arrow; passing the reference avoids an extra arrow function (and an extra `v8` coverage point).
- The default-case branch in `refusalMessage` is unreachable given `Exclude<InstallKind, "global">`, but ESLint requires it; suppressed with `/* v8 ignore next 3 @preserve */` matching the codebase's convention in `ticketDoctor.ts`.

### DIFF
--- a/src/cli.test.ts
+++ b/src/cli.test.ts
@@ -10,6 +10,12 @@ import { resumeWorkspaceCli } from "./commands/resumeWorkspace.ts";
 import { setupReposCli } from "./commands/setupRepos.ts";
 import { setupWorkspaceCli } from "./commands/setupWorkspace.ts";
 import {
+  createDefaultUpgradeCliOptions,
+  upgradeCli,
+  type UpgradeCliOptions,
+} from "./commands/upgrade.ts";
+import { computeUpgradeNudge } from "./lib/upgrade.ts";
+import {
   captureConsoleError,
   captureConsoleLog,
   type ConsoleCapture,
@@ -36,6 +42,17 @@ vi.mock(import("./commands/setupWorkspace.ts"), () => ({
 vi.mock(import("./commands/setupRepos.ts"), () => ({
   setupReposCli: vi.fn<typeof setupReposCli>(),
 }));
+vi.mock(import("./commands/upgrade.ts"), () => ({
+  upgradeCli: vi.fn<typeof upgradeCli>(),
+  createDefaultUpgradeCliOptions: vi.fn<typeof createDefaultUpgradeCliOptions>(),
+}));
+vi.mock(import("./lib/upgrade.ts"), async (importOriginal) => {
+  const actual = await importOriginal();
+  return {
+    ...actual,
+    computeUpgradeNudge: vi.fn<typeof computeUpgradeNudge>(),
+  };
+});
 
 const orchestrateMock = vi.mocked(orchestrate);
 const doctorMock = vi.mocked(doctor);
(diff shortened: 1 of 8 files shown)


paulbaranowski:
## Fixes Applied Successfully

Fixed 3 file(s) based on 3 CodeRabbit feedback item(s) — all on the same theme: opportunistic cache writes in load-bearing positions could fail the user's command on disk errors.

**Files modified:**
- `src/cli.ts` — wrap the passive nudge call in try/catch
- `src/commands/upgrade.ts` — wrap cache prime in `fetchOrFail`
- `src/lib/upgrade.ts` — wrap cache prime in `computeUpgradeNudge` so a freshly-fetched value is never discarded by a disk error

Each fix has a matching test (one new test per file) to keep the 100% coverage gate green.

**Commit:** `fa818c6`

The latest autofix changes are on the `emdash/gc-update-1wxec` branch.

PR #207, opened 2026-06-09
feat: load the initial prompt from a file (prompts.promptFile)

paulbaranowski:
## Why

Storing a large agent prompt inline is poor ergonomics. The `.ts` config solved it by `readFileSync`-ing a sibling `prompt-initial.md` into `prompts.initial` at module load — but the groundcrew-config TUI always saves `crew.config.json`, which inlines the entire prompt (7KB+) as one JSON string and loses the separate-file workflow. JSON cannot reference an external file, so this adds a native loader feature.

## What

Add `prompts.promptFile` to groundcrew's config: a path to a UTF-8 file whose contents become the initial prompt (`prompts.initial` in the resolved config).

- **Resolution:** relative to the config file's directory (matching the `.ts` `import.meta.url` behavior); `~` is expanded; absolute paths are used as-is.
- **Mutual exclusion:** `prompts.initial` and `prompts.promptFile` are mutually exclusive — a hard error if both are set. Neither → the built-in `DEFAULT_PROMPT_INITIAL`.
- **Read at load time;** existing placeholder validation runs on the loaded contents (an unknown `{{placeholder}}` in the file fails exactly as it would inline).
- **Zero blast radius:** `ResolvedConfig.prompts.initial` stays a resolved `string`, so the orchestrator, launch, and rendering are all unchanged. This is a pure loader feature.
- **Backward compatible:** existing inline `prompts.initial` configs keep working.

All changes are in `src/lib/config.ts`: a `resolveInitialPrompt()` helper, threading a `configDir` into `applyDefaults`, and passing `path.dirname(filepath)` at the single call site in `loadConfigWithSource`.

## Tests

New focused suite `src/lib/config.promptFile.test.ts` (split out to stay under the 2000-line `max-lines` cap on `config.test.ts`):

- loads file contents into `prompts.initial`
- relative path resolves against the config dir, **not** cwd (decoy file in cwd proves it)
- `~` expansion and absolute paths
- both `initial` + `promptFile` set → rejected
- unreadable file → error includes the resolved path
- empty / non-string `promptFile` → rejected
- placeholder validation runs on file contents
- zero-placeholder file loads cleanly
- inline `prompts.initial` regression

`node --run verify` passes: all tests green, **100% coverage** (statements/branches/functions/lines), lint/format/cpd/spell/architecture all clean.

## Docs

- `docs/configuration.md`: new "Loading the prompt from a file" section + Full Reference row.
- `crew.config.example.ts`: commented `promptFile` example.

## Decisions

- **No manual version bump.** The repo's semantic-release automation handles the minor publish from this `feat:` commit on merge, so `package.json` is left untouched.
- **Reverted an unrelated `package-lock.json` version-field sync** (pre-existing 4.24.1→4.26.1 drift that `npm install` surfaced) to keep this PR scoped to the feature.
- **Error message wording** keeps the existing `prompts.initial …` text from the shared placeholder validator (the file's contents *become* `prompts.initial`), rather than introducing a `promptFile`-specific variant.

## Scope

This is **Phase 1** (groundcrew engine) of a 3-phase plan. Phase 2 (groundcrew-config TUI wiring) and Phase 3 (migrating an existing config) live in other repos and are explicitly blocked on a published Phase 1.

### DIFF
--- a/crew.config.example.ts
+++ b/crew.config.example.ts
@@ -120,6 +120,9 @@ export default {
   //   // `${XDG_CONFIG_HOME:-$HOME/.config}/groundcrew/initial-prompt.md`.
   //   // If you uncomment this, also uncomment the readFileSync import above.
   //   initial: readFileSync(new URL("./initial-prompt.md", import.meta.url), "utf8"),
+  //   // Or, instead of `initial`, point at a file (also works in crew.config.json).
+  //   // Resolved relative to this config's directory; `~` and absolute paths work.
+  //   // promptFile: "initial-prompt.md",
   // },
   //
   // // Terminal session manager. "auto" picks cmux when on PATH, else tmux.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -157,6 +157,22 @@ export default {
 
 This keeps package defaults portable while letting your private config reference team-specific statuses, tools, plugins, or review loops.
 
+### Loading the prompt from a file (`prompts.promptFile`)
+
+`readFileSync` works for `.ts` configs but not for `crew.config.json`, which has no way to reference an external file. Set `prompts.promptFile` instead and groundcrew reads the file's contents into the initial prompt at load time:
+
+```json
+{
+  "prompts": {
+    "promptFile": "prompt-initial.md"
+  }
+}
+```
+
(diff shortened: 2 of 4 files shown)


therockstorm:
@coderabbitai review

therockstorm:
@mendral-app review

PR #210, opened 2026-06-09
feat: add first-class safehouse.gitSshToHttps config

paulb-instacart:
## Problem

groundcrew's safehouse runner sandboxes agents with `sandbox-exec`, which blocks raw SSH egress (port 22) — only HTTPS through the clearance proxy is reachable. But worktrees (e.g. graft sparse-checkouts of an SSH monorepo) carry SSH remotes, so any `git fetch`/`push` an agent runs inside the sandbox fails.

The workaround was to smuggle a raw `env GIT_CONFIG_* …` prefix into an agent's `cmd` string to serve those SSH remotes over HTTPS. That was:

- **Opaque** — a wall of `GIT_CONFIG_COUNT`/`KEY`/`VALUE` tokens nobody could re-derive by hand or a config UI could introspect.
- **Fragile** — the HTTPS remotes it produces only authenticate via the agent's `gh` credential helper, which reads `GITHUB_TOKEN`. So it had to be paired with a separate `preLaunchEnv: ["GITHUB_TOKEN"]` entry, in a different place. Set it without the token and pushes failed *silently* inside the sandbox, mid-run, long after launch.

```ts
// the old way — two coupled settings you just had to know were connected
claude: {
  cmd: "env GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=url.https://github.com/.insteadOf GIT_CONFIG_VALUE_0=git@github.com: claude --permission-mode bypassPermissions",
  preLaunchEnv: ["GITHUB_TOKEN"],
}
```

## Solution

A first-class, typed `safehouse` config block with a `gitSshToHttps: "enabled"` toggle:

```ts
// crew.config.ts
export default {
  workspace: {
    projectDir: "~/dev",
    knownRepositories: ["my-repo"],
  },
  agents: {
    definitions: {
      claude: {
        cmd: "claude --permission-mode bypassPermissions",
        safehouse: {
          gitSshToHttps: "enabled",
        },
      },
    },
  },
};
```

It's an opt-in toggle: **omit it (default) and git is left untouched** (SSH remotes still fail in the sandbox); **set `"enabled"`** and groundcrew serves the agent's SSH remotes over HTTPS. What that does under the hood:

- **Same additive mapping, composed by groundcrew.** `gitSshToHttps` expands to the same `GIT_CONFIG_COUNT`/`KEY`/`VALUE` layer onto the agent wrap — additive (not `GIT_CONFIG_GLOBAL`), so the agent's global git config and `gh` credential helper survive. The profile shim still keys on the bare agent name, so the emitted launch command is byte-identical to the old hand-written prefix.
- **Self-contained — no coupling to remember.** It auto-forwards `GITHUB_TOKEN` into the agent wrap, deduped against any explicit `preLaunchEnv` entry. No separate token pairing required. (Safe to forward: `GITHUB_TOKEN` is not a build secret, so it isn't scrubbed between the prepare and agent wraps.)
- **Fail-fast at load time.** If `gitSshToHttps` is `"enabled"` but `GITHUB_TOKEN` is unset (or empty) in the environment, the config load **throws** with a clear message — converting the old silent, mid-run push failure into an immediate, actionable error before any agent launches.

Scope is `gitSshToHttps` only. The companion `addDirs` (extra filesystem grants) field is intentionally deferred — a directory-granting field that silently did nothing would be a security footgun.

### Safety notes

- **Safe on repos already using HTTPS.** git's `insteadOf` is a prefix-match: it only rewrites URLs beginning with `git@github.com:`. An already-HTTPS remote doesn't match and passes through untouched; non-GitHub remotes are left alone. (Caveat: it matches the scp-like `git@github.com:` form, not `ssh://git@github.com/`.)
- **String enum, not a bare boolean.** `gitSshToHttps` is a closed enum (`"enabled"`) to match the config's `auto`/enum house style and leave room for future modes (e.g. `"disabled"`/`"auto"`) without a breaking rename. It is GitHub-only today; the name is host-agnostic so adding GitLab/Bitbucket later is a capability expansion, not a rename.

## Why is `safehouse` under `agents`, not top-level or per-repo?

There are deliberately two safehouse-related knobs at different levels:

| Knob | Level | Answers |
| --- | --- | --- |
| `local.runner` (`safehouse` / `sdx` / `none` / `auto`) | top-level | "Am I sandboxed at all, and by which backend?" |
| `agents.definitions.<name>.safehouse.gitSshToHttps` | per-agent | "Inside that sandbox, how does *this agent's* git/credentials behave?" |

The global "which sandbox" decision already *is* top-level (`local.runner`); per-agent `isolation` and `agents.isolation` were both removed to consolidate that choice. The `safehouse` block is the per-launch tuning that rides on top.

**Why `gitSshToHttps` is per-agent:**

- It's mechanically a property of the **agent's launch command** — it expands into an `env GIT_CONFIG_*` prefix on the agent wrap and auto-forwards `GITHUB_TOKEN` into *that agent's* `--env-pass`. The launch builder is handed an `AgentDefinition`, so that's the object it attaches to. It's also the typed replacement for what was literally a per-agent `cmd` prefix (same scope, byte-identical output).
- **Least privilege.** It injects a `GITHUB_TOKEN` into the agent. A read-only reviewer agent shouldn't get a token just because an implementer agent needs to push. A top-level setting would force it onto every agent. It also sits naturally alongside its per-agent siblings (`cmd`, `preLaunch`, `preLaunchEnv`, `sandbox`).

**Why not per-repo (a reasonable alternative):** conceptually the *cause* (SSH remotes the sandbox can't reach) is a property of the repository. But two things make per-agent the right first step today: (1) repository recipes carry only clone/provision concerns (`projectDirOverride`, `provision`, `workdir`) and never flow into the launch-command builder, so per-repo would need new dispatch-time `(repo, agent)` plumbing rather than a field move; and (2) a single agent definition is repo-agnostic — the same agent is dispatched against any known repo, so the setting has to attach to something present at launch. Per-repo remains a legitimate future shape if it proves to be repo-driven in practice (on for the SSH monorepo, off elsewhere) — the only per-repo cost worth scoping is the token forward, since the rewrite itself is a safe no-op on already-HTTPS/non-GitHub remotes.

## Validation

- `node --run verify` — all stages pass: `architecture:check`, `build`, `cpd`, `format:check`, `knip`, `lint`, `markdown:lint`, `spell:check`, `syncpack:lint`, and `test` at 100% coverage (1636 tests).
- TDD red → green throughout: config schema (merge / empty-block / reject-unknown / reject-non-object / reject-when-GITHUB_TOKEN-unset / reject-when-GITHUB_TOKEN-empty) in `config.test.ts`, and the GIT_CONFIG prefix injection + byte-equivalence + GITHUB_TOKEN auto-forward tests in `launchCommand.test.ts`.

## Notes

- Pushed from fork `paulb-instacart/groundcrew`; PR targets `ClipboardHealth/groundcrew:main`.

### DIFF
--- a/src/lib/config.test.ts
+++ b/src/lib/config.test.ts
@@ -798,24 +798,24 @@ describe("loadConfig", () => {
     expect(actual.agents.definitions["claude"]?.color).toBeTypeOf("string");
   });
 
-  it("merges safehouse.gitRewrite through an override and preserves cmd/color defaults", async () => {
+  it("merges safehouse.gitSshToHttps through an override and preserves cmd/color defaults", async () => {
     const configPath = writeConfigFile(
       temporary,
       configSource({
         workspace: VALID_WORKSPACE(temporary),
-        agents: { definitions: { claude: { safehouse: { gitRewrite: "ssh-to-https" } } } },
+        agents: { definitions: { claude: { safehouse: { gitSshToHttps: "enabled" } } } },
       }),
     );
     setEnvironmentVariable("GROUNDCREW_CONFIG", configPath);
     setEnvironmentVariable("GITHUB_TOKEN", "ghp_test_token");
     const { loadConfig } = await loadFreshConfig();
     const actual = await loadConfig();
-    expect(actual.agents.definitions["claude"]?.safehouse?.gitRewrite).toBe("ssh-to-https");
+    expect(actual.agents.definitions["claude"]?.safehouse?.gitSshToHttps).toBe("enabled");
     expect(actual.agents.definitions["claude"]?.cmd).toBeTypeOf("string");
     expect(actual.agents.definitions["claude"]?.color).toBeTypeOf("string");
   });
 
-  it("accepts an empty safehouse block and leaves gitRewrite unset", async () => {
+  it("accepts an empty safehouse block and leaves gitSshToHttps unset", async () => {
     const configPath = writeConfigFile(
       temporary,
       configSource({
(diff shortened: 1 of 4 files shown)


therockstorm:
@paulbaranowski @paulb-instacart Have you tried this config? My coworkers and I are using `gh` from the `safehouse` + `clearance` sandbox.

```
[credential "https://github.com"]
	helper =
	helper = !/opt/homebrew/bin/gh auth git-credential
[credential "https://gist.github.com"]
	helper =
	helper = !/opt/homebrew/bin/gh auth git-credential
[url "https://github.com/"]
	insteadOf = git@github.com:
```

paulbaranowski:
I will try it out and see if it works

therockstorm:
Going to close. We can always re-open if the config change above doesn't address it.

PR #215, opened 2026-06-10
feat: grant safehouse git access for graft and workdir worktrees

paulb-instacart:
## Problem

The safehouse sandbox only auto-grants the agent's working directory (its cwd). That breaks `git` for two worktree shapes groundcrew creates:

- **`workdir` subprojects** — the agent cwd's into a subdirectory of the checkout, so the cwd grant never reaches the worktree-root `.git` gitfile. Git can't find the repo.
- **graft / sparse-checkout worktrees** — the object store lives **outside** the worktree tree (e.g. `~/carrot/.git`), so even granting the whole worktree root isn't enough; the real store is unreachable and git fails with "Operation not permitted".

Separately, `crew doctor` mis-probes agent commands: a `cmd` like `env GIT_CONFIG_COUNT=1 … claude` made doctor treat `GIT_CONFIG_COUNT=1` as a PATH binary and report it missing.

## Solution

Resolve the real filesystem grant set in `composeAgentLaunch` — `[worktreeDir, gitCommonDir]`, deduped — and thread it through `buildLaunchCommand` as `--add-dirs` on **both** safehouse wraps (the `prepareWorktree` hook and the agent), so git works in the hook and the agent alike. The git common dir is resolved from the worktree itself via `git rev-parse --git-common-dir --path-format=absolute`, so an external/relocated store is covered instead of assuming `<projectDir>/<repo>/.git`.

For doctor, reuse the exported `isEnvironmentAssignment` to skip `env VAR=val` assignment tokens, mirroring the launch path's `inferAgentCommandName` (so `env GIT_CONFIG_*=… claude` resolves to `claude`).

## Security: is this a hole?

Short answer: **no — it's a deliberate, scoped widening of the filesystem sandbox, not a new escape vector.** Specifics:

- **No new network egress.** This change touches only filesystem `--add-dirs`; the clearance proxy's egress allowlist is untouched. It opens no new exfiltration or fetch channel.
- **No host secrets exposed.** The grants do **not** include the host home, SSH private keys, credential stores, system paths, or the workspace parent directory.
- **Paths come from trusted config/git metadata, not ticket input.** `worktreeDir` is an operator-controlled checkout path and `gitCommonDir` comes from `git` itself — neither is derived from untrusted ticket/PR content, so there's no path-injection vector.
- **Scope of the widening.** `worktreeDir` and `gitCommonDir` are the agent's own repo surface and the minimum Safehouse needs for Git to work in these worktree shapes. The broader `projectDir` grant was intentionally removed so sibling repos/worktrees are not write-granted.

## Validation

- `node --run verify` — green: 60 files, 1624 tests, 100% statement/branch/function/line coverage (lint, type-check, format, knip, cpd, spell all pass).
- New unit tests: `--add-dirs` appears once per safehouse wrap and is omitted when no grants are requested (`launchCommand.test.ts`); both safehouse-launch assertions in `setupWorkspace.test.ts` expect the grant flag; doctor skips `env KEY=val` and checks the wrapped command (`doctor.test.ts`).

## Notes

- This PR only opens the filesystem grants git needs. The SSH→HTTPS git rewrite for the agent sandbox is handled separately by a followup — #210 (`feat: add first-class safehouse.gitRewrite config`), which promotes it to a first-class `safehouse.gitRewrite` config field rather than smuggling it through a per-agent `cmd` string.
- srt already fences its equivalent surface (worktree root + git common dir) through its settings file; this brings the safehouse runner's git access in line for the worktree shapes above.

Agent session: `codex resume 019eb347-0f21-74f3-a755-e504cd538a90`

<sub>🤖 <code>commit-push-pr:created v1 core@3.7.1</code></sub>

### DIFF
--- a/config/cspell.json
+++ b/config/cspell.json
@@ -50,6 +50,7 @@
     "fsmonitor",
     "gitdir",
     "gitdirs",
+    "gitfile",
     "groundcrew",
     "defence",
     "discriminable",
--- a/src/commands/doctor.test.ts
+++ b/src/commands/doctor.test.ts
@@ -411,6 +411,30 @@ describe(doctor, () => {
     expect(checked).not.toContain("script.ts");
   });
 
+  it("skips `env KEY=val` assignments and checks the wrapped command", async () => {
+    loadConfigMock.mockResolvedValue(
+      makeConfig({
+        default: "env-cli",
+        definitions: {
+          // An `env VAR=val …` prefix (e.g. the safehouse SSH→HTTPS git rewrite)
+          // must not make doctor probe the assignment as a PATH binary. Mirrors
+          // the launch path's inferAgentCommandName, which skips assignments.
+          "env-cli": {
+            cmd: "env GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=url.https://github.com/.insteadOf claude --permission-mode auto",
+            color: "#fff",
+          },
+        },
+      }),
+    );
+
+    await doctor();
+
+    const checked = checkedCommands();
+    expect(checked).toContain("env");
+    expect(checked).toContain("claude");
+    expect(checked).not.toContain("GIT_CONFIG_COUNT=1");
+  });
+
   it("does not probe a built-in agent that is not enabled", async () => {
     // The default makeConfig fixture has only `claude` in `definitions`.
     // `gatherToolTokens` iterates `Object.values(definitions)`, so codex is
--- a/src/commands/doctor.ts
+++ b/src/commands/doctor.ts
@@ -16,6 +16,7 @@ import {
   worktreeBaseDir,
 } from "../lib/config.ts";
 import { detectHostCapabilities, type HostCapabilities, which } from "../lib/host.ts";
+import { isEnvironmentAssignment } from "../lib/launchCommand.ts";
 import { resolveLocalRunner } from "../lib/localRunner.ts";
 import { gatedAgents } from "../lib/usage.ts";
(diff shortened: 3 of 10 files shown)


therockstorm:
@mendral-app review

therockstorm:
@coderabbitai review

therockstorm:
## Review-body findings

**Already fixed**

- Mendral LGTM reviewed the original broad grant and mentioned `projectDir` as the broadest grant. That was superseded by `f9d4c4aa`, which removes `projectDir` from the Safehouse grant set and refreshes the PR body/tests accordingly.

**Disagree**

- CodeRabbit on `src/lib/agentLaunch.ts:63-81`: leaving the current comment. It now describes only `worktreeDir` and the resolved git common dir; the remaining `<projectDir>/<repo>/.git` text describes the old native-layout assumption, not a granted directory. The `resolveGitCommonDir` failure path is already the normal pre-launch exception path: `runCommand` throws, and setup/resume cleanup happens in the existing caller `try`/`catch`, so adding that call-stack detail to this helper comment would be extra narration rather than a behavioral clarification.

## Conversation-tab comments

**Already fixed**

- Mendral review trigger comment was fulfilled by the Mendral review body assessed above.
- CodeRabbit review trigger comment was fulfilled by the CodeRabbit review body assessed above.

```fingerprints
c8fe558087f85d9d
7fda0c1eb940f759
0bda1105eb15b2a9
6904effcdc889e50
```

<sub>🤖 <code>babysit-pr:addressed v1 core@3.7.1</code></sub>

therockstorm (approved):

PR #216, opened 2026-06-10
fix: allow crew resume into a restored-but-empty cmux workspace

Davidster:
## Why

`crew resume <task>` was incorrectly blocked after a machine reboot when the cmux workspace backend is in use.

cmux persists and restores workspace tabs across a reboot, but the agent process inside a restored tab is dead. The cmux probe mapped **every** restored workspace title to a live workspace (`src/lib/cmuxAdapter.ts:53`), so `failIfWorkspaceAlreadyLive` rejected the resume with `Workspace for <task> is already live; attach to it instead of resuming.` — even though nothing was running. Discovered via `op-crew resume tg-3807` failing after a reboot.

`cmux list-workspaces --json` exposes no per-workspace **local**-liveness signal — `remote.active_terminal_sessions` covers SSH sessions only — so there is nothing in the JSON to key on directly.

## What

**Boot-epoch liveness markers in the cmux adapter.** `open` stamps a per-workspace marker file with the host boot epoch (derived from `os.uptime()`, so it's cross-platform and subprocess-free). `list` compares it: a marker that predates the current boot means the machine rebooted since the agent launched, so the workspace is reported `state: "exited"` per the `WorkspaceProbe` contract. A missing or unparseable marker is treated as **live**, preserving the prior behavior and never freeing a running agent's slot. The marker lives under `tmpdir` (the macOS launchd cleaner leaves it intact across reboot, so the pre-reboot epoch survives to be compared) and is cleared on `close`.

This mirrors the existing **zellij exit-marker** pattern, keyed to boot identity rather than command exit. tmux and zellij are unaffected because neither restores tabs across a reboot — they expose real session liveness.

**Resume now blocks only on genuinely-live workspaces** (`src/commands/resumeWorkspace.ts`): present in `names` **and not** in `exitedNames`. A real running agent still prevents a duplicate; a restored-empty tab re-hydrates the worktree and relaunches.

### Why a marker rather than a cmux JSON field

The task notes `remote.active_terminal_sessions` is the only liveness-looking field and it's SSH-only. Rather than depend on an unverified/unstable cmux-internal field, the adapter owns an out-of-band liveness signal — the same design choice the zellij adapter already makes for a backend with no headless exit-state.

## Areas of concern

- Relies on `tmpdir` surviving a reboot on macOS (it does; only periodically cleaned). If a marker is cleaned away, the workspace falls back to "live" (current behavior) — conservative, never frees a running slot.
- A restored-dead cmux tab is left in place when resume relaunches (cmux titles aren't unique handles); the next `open` overwrites the marker. Closing the stale tab was deliberately left out to keep the change minimal, matching the task's accepted "relaunch into the existing workspace" outcome.

## Tested

- New unit tests in `src/lib/workspaces.test.ts` (cmux probe): reboot→exited, marker-matches→live, no-marker→live, corrupt-marker→live, and best-effort marker-write failure on `open`.
- New test in `src/commands/resumeWorkspace.test.ts`: resume proceeds for a probe-reported `exited` workspace; existing "already live" test still guards the live case.
- `node --run verify` passes — all checks green, 100% coverage (statements/branches/functions/lines).

Closes tg-3836

### DIFF
--- a/src/commands/resumeWorkspace.test.ts
+++ b/src/commands/resumeWorkspace.test.ts
@@ -443,6 +443,24 @@ describe(resumeWorkspace, () => {
     expect(workspacesOpenMock).not.toHaveBeenCalled();
   });
 
+  it("proceeds when the workspace was restored after a reboot and reports as exited", async () => {
+    // A cmux tab cmux re-painted after a reboot: present in `names` but flagged
+    // exited because no agent is running inside it. Resume must relaunch.
+    workspacesProbeMock.mockResolvedValue({
+      kind: "ok",
+      names: new Set(["team-1"]),
+      exitedNames: new Set(["team-1"]),
+    });
+
+    await resumeWorkspace(config, { task: "team-1" });
+
+    expect(workspacesOpenMock).toHaveBeenCalledWith(
+      config,
+      expect.objectContaining({ name: "team-1" }),
+    );
+    expect(lastRecordedRunState()).toMatchObject({ task: "team-1", state: "resumed" });
+  });
+
   it("fails closed when workspace liveness cannot be verified", async () => {
     workspacesProbeMock.mockResolvedValue({ kind: "unavailable" });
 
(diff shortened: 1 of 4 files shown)


Davidster:
Closing: scope grew too large to take on right now. The cmux restored-after-reboot fix is committed on branch davidhuculak-tg-3836 (boot-epoch liveness markers) if we revisit, but `node --run verify` is gated by a pre-existing, unrelated coverage-collection flake (parallel V8 coverage contention in verifyAll.mts — reproduces 4/6 on main). Closing this PR and the TG-3836 ticket with it.

PR #219, opened 2026-06-11
feat: hour units for todo-txt rec: recurrence

therockstorm:
## Summary

`rec:Nh` advances a `t:` threshold by N hours on completion — the second half of sub-day recurring tasks (datetime thresholds landed in #218). Design choices:

- **Non-strict `rec:2h` advances from the completion instant** (source-timezone wall clock), not the previous `t:`. This is the stampede guard: after daemon downtime, the task re-arms N hours from now rather than burning through every missed slot in quick succession. **Strict `rec:+2h`** keeps schedule-aligned advancement from the previous threshold, mirroring `due:`'s documented strict semantics.
- **Hourly requires `t:` and is incompatible with `due:`** — `verify()` enforces it; a line that bypassed verify degrades gracefully (due: carried forward unchanged) instead of feeding hours into date-only math.
- **Same-day id evolution fixed:** `advanceId` previously appended the (unchanged) compact date for same-date recurrence — `sweep-20260608-20260608`, growing each cycle. It now reuses the base id and lets `buildUniqueId` number cycles (-002, -003, …); the existing date-advance path is byte-identical.

## Validation

- TDD, red first: 8 new tests — non-strict from-completion, strict from-previous (datetime, seconds-bearing, and date-only-as-midnight forms), midnight crossing advances the id date, source-timezone wall-clock math (America/Chicago), verify accepts `rec:2h`+`t:`, rejects hourly without `t:` and hourly+`due:`, and the due-carry-forward degradation.
- `node --run verify` fully green; coverage 100% (2884/2884 branches).

## Notes

- Wall-clock hour math ignores DST transitions deliberately (a sweep at 2h cadence doesn't care); documented in `advanceThreshold`.
- Deployment: restart the daemon after merging, then the recurring flaky-triage todo line switches to `rec:2h` with a datetime `t:` (already staged in ~/v/todo.md).

Agent session: `claude --resume b2a3a577-12d9-4bd3-bf10-1c16957f6da4`

<sub>🤖 <code>cb-ship:created v1 core@3.8.0</code></sub>

### DIFF
--- a/config/cspell.json
+++ b/config/cspell.json
@@ -137,6 +137,7 @@
     "sdlc",
     "tspan",
     "dwmy",
+    "dwmyh",
     "recurrence",
     "todopath",
     "todotxt",
--- a/src/lib/adapters/todo-txt/source.test.ts
+++ b/src/lib/adapters/todo-txt/source.test.ts
@@ -215,6 +215,24 @@ describe("TodoTxtTaskSource", () => {
     expect(readFileSync(path.join(tmp.tasksDir, "PROMPT-1.md"), "utf8")).toBe("Existing prompt.\n");
   });
 
+  it("createTask accepts hourly recurrence", async () => {
+    tmp.writeTodo("");
+
+    const source = makeSource(tmp);
+    await source.createTask?.({
+      title: "Hourly sweep",
+      agent: "claude",
+      id: "SWEEP-1",
+      projects: [],
+      contexts: [],
+      dependencies: [],
+      edit: false,
+      recurrence: "2h",
+    });
+
+    expect(readFileSync(tmp.todoPath, "utf8")).toContain("rec:2h status:todo");
+  });
+
   it("creates the todo file when it does not exist", async () => {
     const source = makeSource(tmp);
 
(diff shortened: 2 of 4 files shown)


therockstorm:
## Review-body findings

### Agree

- **CodeRabbit (outside-diff, source.ts:21): `createTask` still rejected hourly recurrence tokens.** Valid — `createTask` kept its own format-only regex excluding `h`, so `crew task create --rec 2h` failed while verify/writeback accepted the unit. Fixed in https://github.com/ClipboardHealth/groundcrew/commit/c63a74c110b18284bd20a6cb37a309ad4695e4c0 (red-first test). Note: `createTask` has no `t:` input today, so an hourly task created via CLI still needs its threshold added by hand before verify passes — acceptable since the CLI also can't set `t:` for any other unit; widening `CreateTaskInput` is separate follow-up work if wanted.

### No action needed

- Mendral LGTM — no findings.

```
4e9bf4b09e722d62
598d3a235d96ed92
```

<sub>🤖 <code>babysit-pr:addressed v1 core@3.8.0</code></sub>

therockstorm:
## Review-body findings

### No action needed

- Mendral LGTM on `c63a74c` (createTask regex sync confirmed) — no findings. CI green across the board.

```
49c416f0107fd80f
```

<sub>🤖 <code>babysit-pr:addressed v1 core@3.8.0</code></sub>

PR #222, opened 2026-06-11
feat: add cursor as a built-in agent preset

ClipboardJason:
## Why

Cursor's `cursor-agent` CLI was only usable as a hand-rolled custom agent (and served as the canonical custom-agent example in tests/docs). This promotes it to a first-class built-in preset at parity with `claude` and `codex`, so `cursor: {}` in `agents.definitions` (or `crew init --agent cursor`) is all that's needed, with usage gating, sandbox policy, and doctor checks included.

## What

- **Preset** (`src/lib/config.ts`): `cursor-agent --sandbox disabled --force`, `#929292`, codexbar provider `cursor`.
  - `--sandbox disabled`: groundcrew sandboxes externally; cursor's own sandbox would nest under Seatbelt and fail (same rationale as codex's bypass flag).
  - Flag order is load-bearing: doctor's tokenizer treats the token after any flag as its value, so the boolean `--force` must come last.
- **srt sandbox policy** (`src/lib/srtPolicy.ts`): codex-style config-home relocation via `CURSOR_CONFIG_DIR`, seeding only `cli-config.json`. Profile is keyed by the command basename `cursor-agent` (how `inferAgentCommandName` resolves it).
  - No read grant on the real `~/.cursor`: relocation makes it unnecessary at runtime (validated live: `cursor-agent status` reports logged-in from a relocated home seeded with only `cli-config.json`), and it holds live third-party secrets (`mcp.json` server `env` blocks), which srt's allowRead-beats-denyRead precedence makes impossible to carve out.
  - `usesMacosKeychain: true`: the bearer token lives in the user keychain (`cli-config.json` carries identity/settings only; safehouse's own `cursor-agent.sb` likewise requires its keychain integration).
  - `mcp.json`/`hooks.json` deliberately not seeded (spawn/exec surfaces — mirrors claude's denied-mcpServers posture). Cross-agent: `.cursor` added to `ALL_AGENT_HOME_DIRS`; the versioned install under `~/.local/share/cursor-agent` is read-only for cursor only and write-denied (host-persistence vector; `.local/bin` only covers the shim).
- **doctor** (`src/commands/doctor.ts`): built-in agent names now map to CLI tokens (`cursor` → `cursor-agent`), fixing the hint logic which assumed name == binary.
- **init**: `crew init --agent cursor`; example config switched to the single-line `// cursor: {},` form init's dedup regex expects.
- **Docs**: README/configuration/runners/troubleshooting updated; also fixed stale `models.*` references the earlier model→agent rename missed (verified each against source).
- **Tests**: fixtures that used `cursor` as the canonical custom agent renamed (`myagent`/`mycli`); new coverage for preset merge, relocation staging, seed exclusions, cross-profile `~/.cursor` isolation, keychain grant, doctor hint, and tokenizer flag-order regression.

## Validation

- `node --run verify` green (full suite).
- `crew doctor` with a cursor-only config: resolves the preset, finds `cursor-agent` on PATH, runner auto→safehouse ready.
- `codexbar usage --provider cursor --source auto --format json` returns live usage windows (gating works without pinning a source).
- Installed safehouse 0.10.1 ships the `cursor-agent.sb` profile, so the macOS auto runner needs no changes.
- Adversarial security review of the sandbox diff (the no-`~/.cursor`-read posture and keychain finding came out of it).
- Live end-to-end Linear dispatch test (agent-cursor label) created https://github.com/ClipboardHealth/groundcrew/pull/224.
image

## Areas of concern

- Linux: `usesMacosKeychain` is a no-op there and the login probe ran on macOS. If cursor-agent on Linux keeps its token in a file under `~/.cursor` other than `cli-config.json`, the relocated home stages logged-out (fails safe, but worth a Linux probe before relying on it).
- Pre-existing, unrelated: `setupWorkspace.test.ts` 'wraps the agent command with Safehouse…' fails on any machine with `NPM_TOKEN`/`BUF_TOKEN` exported — the test reads real `process.env`, so the staged `--env-pass=NPM_TOKEN,BUF_TOKEN` flag breaks its adjacency assertion. Reproduces on clean main; happy to file separately.


ClipboardJason:
Closing this in favour of the custom configurations. Not much point to this.

PR #237, opened 2026-06-12
feat: add opt-in tmux session-per-task workspace model

chandanand:
## Summary

The tmux workspace backend still defaults to the current window model: every task opens as a window inside one shared `groundcrew` tmux session. This PR keeps the session-per-task model as an env-only opt-in with `GROUNDCREW_TMUX_SESSION_PER_TASK=1`; there is no `crew.config.ts` surface for the mode.

When the resolved workspace backend is tmux and the env opt-in is not set, `crew` prints a startup warning that session-per-task mode will become the default soon. That warning is emitted only from the tmux adapter path, so auto users who resolve to cmux and explicit zellij users do not see it.

Migration plan for tmux users:

- Opt in with `GROUNDCREW_TMUX_SESSION_PER_TASK=1`.
- Attach with `tmux attach -t <task>` instead of `tmux attach -t groundcrew:<task>`.
- Each task gets its own managed tmux session, so windows and panes are task-local.
- `crew stop` and `crew cleanup` close the managed session. Untagged same-name user sessions are ignored.
- Finished sessions disappear unless `GROUNDCREW_KEEP_DEAD_WINDOWS=1` keeps exited scrollback.

## Validation

- `node --run verify` - full suite passes, 64 test files, 1745 tests, 100% statements/branches/functions/lines coverage.
- Added regression coverage that cmux-resolved auto users and zellij users do not receive the tmux migration warning.

<sub>🤖 <code>commit-push-pr:created v1 core@3.7.1</code></sub>
<sub>🤖 <code>cb-work:updated v1</code></sub>

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -17,7 +17,7 @@ Branch name is `<os-username>-<task-lowercased>`. One task can have at most one 
 The host-side terminal session that runs an agent for one task. Two kinds, one concept:
 
 - **cmux workspace** — a pane/tab in [cmux](https://github.com/clayton-cole/cmux).
-- **tmux workspace** — a window inside a dedicated `groundcrew` tmux session.
+- **tmux workspace** — by default a window inside a shared `groundcrew` tmux session; set `tmux.perTaskMode: "session"` (or the `GROUNDCREW_TMUX_SESSION_PER_TASK=1` env override) for a dedicated session per task (tagged `@groundcrew_managed`) so each task isolates its own windows/panes.
 
 Every provisioned task gets one workspace, named with the task id (`TEAM-220`). Tracked by task — one workspace per task.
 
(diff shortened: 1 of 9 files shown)


chandanand:
## Review-body findings

### Disagree

- **`accessHint` shell injection via `spec.name` (`src/lib/tmuxAdapter.ts:58`)** — Leaving current behavior. `accessHint` returns a **display-only** string (rendered in `crew status` output and setup prompts); groundcrew never shell-executes it — every real tmux invocation goes through `runWorkspaceCommand("tmux", [...argv])`, which passes an argv array and bypasses the shell entirely, so there is no command-execution surface. Additionally, task ids are validated to `[a-z0-9-]` by `PLAIN_TASK_ID_RE` (`src/lib/taskId.ts`) and asserted at worktree creation (`src/lib/worktrees.ts:183`); a name containing a space, `;`, backtick, or `:` cannot reach `accessHint`, and tmux itself rejects such session names. The displayed `tmux attach -t <id>` is therefore always safe. No change made.

```fingerprints
ee2d1abb0a70c6dd
```

<sub>🤖 <code>babysit-pr:addressed v1 core@3.7.1</code></sub>

therockstorm (approved):
Thanks!

PR #251, opened 2026-06-15
fix: reuse an existing branch when dispatching a watched task

ClipboardJason:
## Why

When `run --watch` dispatches a task it believes is new, it can hit a leftover git branch from a prior run and **hard-fail every tick** instead of reusing the existing work.

- Eligibility (`classifyRecovery`) treats a task as already-provisioned only when its **worktree directory** is found on disk. When the directory is missing it's classified as a fresh start.
- But `removeWorktree` deletes the git branch only **best-effort** (`deleteBranchBestEffort` swallows failures), and operators sometimes `rm -rf` the directory by hand. So it's routine to end up with the **directory gone but the branch still present**.
- On the next tick `createWorktree` runs `git worktree add -b <branch> <dir> origin/main`, which fails with `fatal: a branch named '<branch>' already exists`. That error isn't a `WorktreeAlreadyExistsError`, so `setupWorkspace` re-throws, the dispatcher catches it and logs a hard failure, and the task is left in Todo to re-fail on every subsequent tick.

## What

`createWorktree` now probes for the branch first (`git branch --list`). When it already exists, it **attaches** the surviving branch into the new worktree (`git worktree add <dir> <branch>` — no `-b`, no fetch), preserving its prior commits rather than crashing. The fresh-branch path is unchanged. Once creation succeeds, `setupWorkspace` opens the workspace and `markInProgress` runs as normal — no changes needed in the dispatcher, eligibility, or `setupWorkspace`.

### Areas of concern / scope boundary

If the lingering branch is *also* already checked out in another (orphaned) worktree, `git worktree add <dir> <branch>` errors with "already checked out". That's a genuine conflict; the dispatcher's existing catch logs it and moves on — same behavior as today, no regression. This PR fixes the dominant "branch lingers, directory gone" case.

## Tests

- Added a test asserting the reuse path attaches the existing branch and skips `-b` / fetch / the default-branch probe.
- The `describe(create)` suite was extracted into `worktrees.create.test.ts` because `worktrees.test.ts` had reached the 2000-line `max-lines` lint cap (any new test there would trip it).

## Validation

`node --run verify` passes locally: 1707 tests across 64 files, branch coverage 100%, lint and types clean.


therockstorm:
## Review-body findings

No actionable findings in the current automated review bodies. Copilot reported no comments, and Mendral reported LGTM.

```cb-babysit:fingerprints
268c2aa5b45927ef
44e539470671ecae
```

<sub>🤖 <code>cb-babysit:addressed v1 core@3.8.0</code></sub>

therockstorm (approved):
Reviewed via cb-babysit: CI is green, no active review feedback, local node --run verify passed, and the change is focused.

PR #274, opened 2026-06-24
feat(config): re-open configurable read-only dirs in the sandbox

pgbezerra:
## What

Adds `local.readOnlyDirs` — a configurable list of host directories re-opened read-only inside the sandbox (both the `safehouse` and `srt` runners). It defaults to `~/.config/tfenv`, so `terraform`/`tfenv` work inside sandboxed agents out of the box.

## Why

The sandbox masks the home region and re-opens common toolchains (nvm, cargo, pyenv, …) but never tfenv's config root. Under the mask an agent running terraform hit `tfenv: No versions available` and `Operation not permitted` reading the version pin.

Rather than special-casing tfenv in the runner code, the re-open list lives in config: tfenv is just the default entry, and anyone whose toolchain lives elsewhere (a relocated tfenv root, or another masked tool entirely) adds or replaces entries without touching code. Read-only so a prompted agent can't mutate host toolchain state.


src/lib/agentLaunch.ts
```diff
@@ -51,6 +53,8 @@ export function composeAgentLaunch(input: {
   omitPromptArgument?: boolean | undefined;
   taskSourceWritePaths?: readonly string[] | undefined;
   safehouseEnableFeatures?: readonly string[] | undefined;
+  /** `config.local.readOnlyDirs`: host dirs re-opened read-only in the sandbox. */
```
pgbezerra: Redundant comment

```suggestion
```

pgbezerra: Addressed in https://github.com/ClipboardHealth/groundcrew/commit/d216a43997ef582b2500bb300f3ddf414052dd48. Removed the redundant doc comment; the sibling fields (`safehouseEnableFeatures`, `taskSourceWritePaths`) carry none either.

<sub>🤖 <code>babysit-pr:addressed v1 core@3.9.0</code></sub>

src/lib/config.ts
```diff
@@ -447,6 +456,36 @@ const DEFAULT_GIT: ResolvedConfig["git"] = {
...
+  };
+}
+
+function normalizeReadOnlyDirs(value: unknown): string[] {
+  const dirs = value ?? DEFAULT_LOCAL_READ_ONLY_DIRS;
+  if (!isStringArray(dirs)) {
+    fail("local.readOnlyDirs must be an array of strings");
+  }
+  return dirs.map(expandHome);
+}
```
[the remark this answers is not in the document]
pgbezerra: Leaving current behavior. The static `~/.config/tfenv` default plus the overridable `local.readOnlyDirs` config field is intentional: this PR deliberately replaced the env-var derivation (`TFENV_CONFIG_DIR`/`XDG_CONFIG_HOME`) with config, so resolution lives in config rather than code. A user whose tfenv root is relocated sets `local.readOnlyDirs` to point there. Note both runners now read the same `config.local.readOnlyDirs`, so safehouse and srt are in sync — the asymmetry flagged here existed only in the earlier env-var version.

<sub>🤖 <code>babysit-pr:addressed v1 core@3.9.0</code></sub>

pgbezerra:
## Review-body findings

### Disagree
- **CodeRabbit — `config.ts` `DEFAULT_LOCAL_READ_ONLY_DIRS` should derive tfenv's root from `TFENV_CONFIG_DIR`/`TFENV_ROOT` instead of hardcoding `~/.config/tfenv`.** Intentional design: this PR deliberately replaced env-var derivation with the overridable `local.readOnlyDirs` config field, so the resolution lives in config rather than code. Users with a relocated tfenv root set `local.readOnlyDirs` to point there. (Replied on the inline thread.)
- **CodeRabbit — `srtPolicy.ts:367` srt still misses relocated tfenv roots / "out of sync with safehouse".** Stale post-refactor: safehouse and srt now both read the same `config.local.readOnlyDirs`, so they are symmetric. The asymmetry described existed only in the earlier env-var version. Same design rationale as above.

### Already fixed
- **Mendral — orphaned JSDoc for `resolveSafehouseAddDirs` (`agentLaunch.ts`).** Resolved during the rebase: `resolveSafehouseReadOnlyDirs` now sits *after* `resolveSafehouseAddDirs` with its own doc block, so neither doc is dangling. Mendral's later review confirms (LGTM).

_Two further Mendral reviews were LGTM with no actionable findings._

```
babysit-pr-fingerprints
80eb843ec61751a9
26b54ec5a6dcf862
e862487564bbd4f4
5bfe121b35ba2484
022f77ada4a34570
```

<sub>🤖 <code>babysit-pr:addressed v1 core@3.9.0</code></sub>

stefanchrobot (approved):

PR #277, opened 2026-06-25
feat(cmux): deliver Claude activity hooks via a project settings file (tg-3925)

ClipboardJason:
## Why

Claude's cmux activity-reporting hooks were injected inline (`claude --settings '<hooks-json>'`), baked into the staged `launch.sh` that crew builds at launch. That delivery is **ephemeral**: when a user exits the agent and restarts Claude by hand in the same worktree — e.g. `claude --resume` to recover from a broken MCP server — the relaunch bypasses `launch.sh` entirely, so the hooks are absent. The cmux sidebar then shows stale progress (frozen at the last reported phase) for the rest of that session.

The cursor agent already avoids this by writing hooks to a project-level file the CLI auto-discovers on every startup. This moves Claude to the same persistent delivery.

## What

- **`cmuxAgentHooks`** gains a per-agent hook-delivery map (`claude` → `.claude/settings.local.json`; codex/cursor → none) and stamps every hook command with an inert trailing shell-comment marker so a re-launch can recognize its own prior entries. `isCmuxAgentHookCommand` is the shared "is this one of our hook commands" check.
- **New `cmuxProjectSettings.writeCmuxAgentProjectSettings`** writes the file into the worktree at launch:
  - git-excluded via the repo's `info/exclude` (idempotently, no duplicate patterns), never committed;
  - skipped entirely when the repo already tracks the file (committed config never clobbered);
  - merged idempotently into any pre-existing `settings.local.json` — unrelated user settings and user-authored hooks are preserved; only prior groundcrew-authored hook entries are replaced.
- **`composeAgentLaunch`** calls the writer (best-effort — a failure is logged and swallowed, never strands a launch) for cmux workspaces, and the inline `--settings` hook injection is dropped so the hooks have a single source and do not double-fire.

The hook events, labels, and `set-progress` commands are unchanged — only the delivery moves. cursor delivery is unchanged; agents with no hook integration (e.g. codex) still receive none.

### Note on runner gating

The old inline injection only fired under the safehouse runner. The project file is now written for any cmux workspace regardless of runner; the file is harmless under runners where `CMUX_WORKSPACE_ID` isn't present (the existing per-hook guard makes each command a no-op), and it's what makes a manual host `claude --resume` keep reporting.

## Validation

- TDD throughout; `node --run verify` green (1979 tests, 100% coverage, lint/format/spell/arch/knip clean).
- **Manual validation still required (left as draft):** launch a Claude task, exit the agent, relaunch in the same worktree via `claude --resume`, and confirm the sidebar progress resumes updating. Per the ticket, if Claude does not load `<worktree>/.claude/settings.local.json` under the safehouse mount, fall back to `.claude/settings.json` (the delivery path is a single constant in `cmuxAgentHooks.ts`).

## Incidental

Fixes a pre-existing lint failure in `teardownReporter.test.ts` (the config `local` literal was missing the recently-added `safehouse`/`readOnlyDirs` fields) so `node --run verify` is green.

Closes tg-3925. Related: TG-3891, TG-3890.

### DIFF
--- a/config/cspell.json
+++ b/config/cspell.json
@@ -58,6 +58,8 @@
     "gitdirs",
     "gitfile",
     "groundcrew",
+    "idempotently",
+    "unmatch",
     "defence",
     "discriminable",
     "initialised",
--- a/src/commands/setupWorkspace.test.ts
+++ b/src/commands/setupWorkspace.test.ts
@@ -818,8 +818,8 @@ describe(setupWorkspace, () => {
     );
     expect(launchScript).toContain('_safehouse_shim="$_safehouse_shim_dir/claude"');
     expect(launchScript).not.toContain("--enable=all-agents");
-    expect(launchScript).toContain("exec claude --permission-mode auto --settings ");
-    expect(launchScript).toContain('"$@"');
+    expect(launchScript).toContain('exec claude --permission-mode auto "$@"');
+    expect(launchScript).not.toContain("--settings ");
     expect(launchScript).toContain('sh "$_p"');
     // prepareWorktree status guard so a failed install still launches the agent
     expect(launchScript).toContain('"$prepare_status" -ne 0');
--- a/src/commands/teardownReporter.test.ts
+++ b/src/commands/teardownReporter.test.ts
@@ -226,7 +226,12 @@ function makeConfig(): ResolvedConfig {
     },
     prompts: { initial: "x" },
     workspaceKind: "auto",
-    local: { runner: "auto", networkEgress: "allowlisted" },
+    local: {
+      runner: "auto",
+      networkEgress: "allowlisted",
+      safehouse: { enable: [] },
+      readOnlyDirs: [],
+    },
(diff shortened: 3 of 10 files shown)


ClipboardJason:
Closing: TG-3925 is being blocked on #260 (TG-3892) rather than implemented off `main`.

This PR was force-started off `main` for dogfooding, but TG-3925's design is to **extend** the per-agent hook-delivery map and project-file writer that the cursor work introduces in #260 (`cursorProjectHooks.ts` / `agentHookDelivery`). Those don't exist on `main` yet, so this branch instead added a parallel `cmuxProjectSettings.ts` — duplicating machinery #260 already provides, which would collide at merge.

TG-3925 is now Planned and `blocked-by` TG-3892. Once #260 lands on main, it'll be re-launched stacked on that so it extends the real delivery map instead of duplicating it. The `jason-tg-3925` branch is preserved here if any of this implementation is worth salvaging.

PR #285, opened 2026-07-03
feat(adapters): enable shell task sources by kind (open registry)

paulbaranowski:
## Why

Today a shell-backed task source has to be enabled with a full inline `{ kind: "shell", commands: {...} }` block. This makes a shell source enable-able by name alone, exactly like a built-in adapter:

```jsonc
sources: [{ kind: "jira" }]   // same shape as { kind: "linear" }
```

groundcrew discovers `source.json` manifests from two roots, turns each into a synthetic adapter whose `kind` is the manifest name, and at enable-time materializes the source's scripts before building a shell `TaskSource`. Enabling a source is also what installs it.

This is **part 2** of a three-part effort ([design](../plans): open-registry). It implements Tasks 2, 3, 4, 5, 6, and 8 of the source plan. Packaging (Task 1) and the unified catalog / discovery-API export (Tasks 7, 9, 10) ship in the sibling PRs.

## What

All under `src/lib/adapters/shell/` unless noted:

- **`manifest.ts`** - internal `sourceManifestSchema` + `SourceManifest` (+ `manifestTimeoutsSchema`). `files` defaults to `[]`.
- **`install.ts`** - `installShellSource({ manifest, manifestDir })`: idempotent copy of `files[]` into the expanded `installDir` (skips when bytes already match), marks each `0o755`, no-op when `files` is empty, never writes secrets.
- **`manifestAdapter.ts`** - `shellConfigFromManifest` + `manifestAdapter`: a synthetic `AdapterDefinition` with `kind = manifest.name` and an override schema `{ kind, name?, env?, timeouts?, enabled? }`; `create` installs the scripts (idempotent) then builds a shell source.
- **`discovery.ts`** - `discoverFromRoots` (pure; later root overrides earlier by name, recording a warning) + `discoverTaskSourceManifests` (package root `<pkg>/task-sources` then user root `${XDG_CONFIG_HOME:-~/.config}/groundcrew/task-sources`; warnings go to stderr).
- **`registry.ts`** - `mergeManifestAdapters` folds discovered manifests into the code-adapter registry, throwing on a reserved-kind collision (`linear`/`shell`/`todo-txt`); `adapterRegistry` now includes discovered sources.
- **`config.ts`** - `SourceConfig` widened with a permissive manifest-source arm.

## Acceptance criteria

- [x] A `source.json` bundle in either root is discovered and exposed as `kind = <manifest name>`.
- [x] `{ kind: "jira" }` builds a working shell task source and materializes the jira script to `installDir` (`enableByKind.test.ts` drives the full chain).
- [x] A user-root source with the same name as a package source wins, and a warning is written to stderr.
- [x] A manifest named `linear`/`shell`/`todo-txt` is rejected with a collision error.
- [x] The generic `{ kind: "shell" }` adapter and its full inline config form are unchanged.
- [x] `node --run verify` is green (architecture:check, build, cpd, format:check, knip, lint, markdown:lint, spell:check, syncpack:lint, test @ 100% coverage).

## Validation

`node --run verify` passes end-to-end. Built via TDD (red → green per task); 36 new tests across `manifest`, `install`, `manifestAdapter`, `discovery`, `registry`, and an end-to-end `enableByKind` integration test. Coverage gate held at 100%.

## Decisions & areas of concern

An independent review flagged six items. How each was handled:

1. **Packaging: `task-sources/` is not yet in `package.json#files`.** Deliberate: shipping the bundle in the published tarball is **Task 1 of the source plan, scoped to the sibling "part 1" PR** (which adds it with a `packaging.test.ts`). In an installed package, package-root discovery depends on that PR landing. Not duplicated here to avoid a conflict. **This PR should merge together with / after part 1.**

2. **A malformed or reserved-name user manifest fails the whole registry (all commands), not just that source.** The reserved-kind collision throwing is required by acceptance criterion 4 and is directly tested. Discovery runs eagerly in the module-level `adapterRegistry`, so a bad user-dropped `source.json` currently propagates. Kept faithful to the plan's strict-validation design; making a single malformed manifest skip-and-warn (while collisions still throw) is a reasonable follow-up if we want per-source isolation.

3. **Registry production-IIFE test was non-hermetic** (asserted the exact key set, which reads the real `~/.config`). **Fixed:** it now asserts the built-ins + `jira` are present rather than an exact set, so a developer's own user-installed sources don't break the suite.

4. **Widened `SourceConfig` accepts any `{ kind: string }`, loosening author-time types.** Intentional per the design (C5): the concrete per-source schema is built at runtime by `manifestAdapter` and validated at `buildSources` time, so the static type is deliberately permissive. Runtime validation of the inline `shell` form is unchanged.

5. **`installShellSource` has no path-traversal containment on `files[]` / `installDir`** (defense-in-depth). Manifests are groundcrew-shipped or user-installed (low trust boundary), so out of scope here; a `path.relative` escape check is a sensible hardening follow-up.

6. **A `files[]` entry missing from the bundle throws a bare `ENOENT`** rather than a source-attributed error. Minor DX follow-up.

Additional robustness added while wiring the registry: `discoverFromRoots` now tolerates an absent or non-directory root (returns no sources) instead of crashing, and re-raises genuinely unexpected read failures. This fixed an unhandled rejection that a broadly-`fs`-mocked test triggered once discovery ran at registry init.

### DIFF
--- a/src/lib/adapters/registry.test.ts
+++ b/src/lib/adapters/registry.test.ts
@@ -12,7 +12,9 @@ import {
   buildRegistry,
   buildSourceConfigSchema,
   listAdapterDirectories,
+  mergeManifestAdapters,
 } from "./registry.ts";
+import type { DiscoveredManifest } from "./shell/discovery.ts";
 
 function emptySource(name: string): TaskSource {
   return {
@@ -122,13 +124,46 @@ describe(listAdapterDirectories, () => {
   });
 });
 
+function discovered(name: string): DiscoveredManifest {
+  return {
+    manifest: {
+      name,
+      kind: "shell",
+      description: name,
+      installDir: "~/.config/groundcrew",
+      files: [],
+      commands: { listTasks: `${name} list` },
+    },
+    manifestDir: `/tmp/${name}`,
+    origin: "user",
+  };
+}
+
+describe(mergeManifestAdapters, () => {
+  it("adds a discovered manifest as an adapter keyed by its name", () => {
+    const code = { shell: fakeAdapter("shell") };
+
+    const merged = mergeManifestAdapters(code, [discovered("jira")]);
+
+    expect(Object.keys(merged).toSorted()).toStrictEqual(["jira", "shell"]);
+    expect(merged["jira"]?.kind).toBe("jira");
+  });
+
+  it("rejects a manifest whose name shadows a built-in adapter kind", () => {
+    const code = { shell: fakeAdapter("shell") };
+
+    expect(() => mergeManifestAdapters(code, [discovered("shell")])).toThrow(/collides/i);
+  });
+});
+
 describe("adapterRegistry production IIFE", () => {
-  it("loads the built-in adapters from src/lib/adapters/", async () => {
+  it("loads the built-in adapters plus discovered manifest sources", async () => {
     const registry = await adapterRegistry;
-    expect(Object.keys(registry).toSorted()).toStrictEqual(["linear", "shell", "todo-txt"]);
+    expect(Object.keys(registry).toSorted()).toStrictEqual(["jira", "linear", "shell", "todo-txt"]);
     expect(registry["linear"]?.kind).toBe("linear");
     expect(registry["shell"]?.kind).toBe("shell");
     expect(registry["todo-txt"]?.kind).toBe("todo-txt");
+    expect(registry["jira"]?.kind).toBe("jira");
   });
 
   it("each loaded adapter exposes a Zod configSchema and a create function", async () => {
--- a/src/lib/adapters/registry.test.ts
+++ b/src/lib/adapters/registry.test.ts
@@ -159,7 +159,12 @@ describe(mergeManifestAdapters, () => {
 describe("adapterRegistry production IIFE", () => {
   it("loads the built-in adapters plus discovered manifest sources", async () => {
     const registry = await adapterRegistry;
-    expect(Object.keys(registry).toSorted()).toStrictEqual(["jira", "linear", "shell", "todo-txt"]);
+    // The three built-in code adapters plus the package-bundled `jira` manifest
+    // source. Asserted by presence rather than exact set so a developer's own
+    // user-installed sources under ~/.config do not break the suite.
+    expect(Object.keys(registry)).toEqual(
+      expect.arrayContaining(["jira", "linear", "shell", "todo-txt"]),
+    );
     expect(registry["linear"]?.kind).toBe("linear");
     expect(registry["shell"]?.kind).toBe("shell");
     expect(registry["todo-txt"]?.kind).toBe("todo-txt");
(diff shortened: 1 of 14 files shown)


paulbaranowski:
Thanks @coderabbitai. Addressed the two Major findings:

- **Stability (discovery.ts): malformed `source.json` fails the whole registry.** `discoverFromRoots` now catches per-manifest JSON/Zod parse errors, records a warning, and continues so every other adapter still loads. Reserved-name collisions still throw in `mergeManifestAdapters` (that hard-error is a required acceptance criterion). Covered by a new invalid-JSON discovery test. (148e19c → 43ce32d)
- **Security (install.ts): `files[]` path traversal.** `installShellSource` now rejects any entry whose resolved path escapes `installDir` (via `path.relative(...).startsWith('..')`) before copy/chmod. Covered by a new `../evil.sh` rejection test.

On the nitpick (duplicated kebab-case name regex): left inline to match the existing convention (`shell/schema.ts` and others already inline the same pattern); `cpd` passes. A shared name-schema constant would be a broader cleanup across those call sites.

`node --run verify` is green (100% coverage held).

therockstorm:
@mendral-app review

therockstorm:
## Review-body findings

**Agree**
- CodeRabbit on `src/lib/sourceCapabilities.ts`: addressed in https://github.com/ClipboardHealth/groundcrew/commit/de46de9d1f1d58bc4976e8f0fe6686327594dfff. Manifest-backed capabilities now reuse a module-level cache, with coverage proving repeated lookups do not rescan discovery.

**Disagree**
- CodeRabbit on duplicated kebab-case validation in `manifestAdapter.ts`: leaving current behavior. The inline validation matches the surrounding schema style, and extracting a shared schema would broaden this PR for a low-value cleanup.

**Already fixed**
- CodeRabbit on malformed manifest discovery: current PR code already catches per-manifest JSON/Zod failures, records a warning, and continues discovery.
- CodeRabbit on install path traversal: current PR code already rejects paths escaping `installDir` before copy/chmod, and the later nested-file fix preserves that guard.

## Conversation-tab comments

**Already fixed**
- paulbaranowski's summary comment: confirmed the referenced major findings are handled; the kebab-case nit remains intentionally unchanged as noted above.
- `@mendral-app review`: Mendral review ran and returned LGTM, so no code action was needed.

```fingerprints
93246081149f8874
5b90f1657e781997
7480c702b9b824d1
0627c733c594e6f5
c8fe558087f85d9d
```

<sub>🤖 <code>cb-babysit:addressed v1 core@3.15.0</code></sub>

therockstorm:
## Review-body findings

**Agree**
- CodeRabbit on `src/lib/sourceCapabilities.test.ts`: addressed in https://github.com/ClipboardHealth/groundcrew/commit/d15668a403dadaaa36d0f2c3d329fb7be4095986. The manifest-backed capability test now imports a fresh module instance before exercising discovery, so it cannot reuse a stale module-level cache.

```fingerprints
93246081149f8874
5b90f1657e781997
7480c702b9b824d1
ecd247b0a5697709
```

<sub>🤖 <code>cb-babysit:addressed v1 core@3.15.0</code></sub>

therockstorm (approved):

PR #294, opened 2026-07-08
feat(adapters): per-OS install hints and editable env defaults for the jira source

paulbaranowski:
This PR closes two manifest-data gaps in the packaged jira task source, both surfaced by crew-config (the TUI). Per-OS install hints are display-only metadata; manifest env defaults feed the packaged source runtime too, so behavior-neutral defaults stay aligned with jira.sh while review/todo pattern defaults seed the setup screen.

## 1. Per-OS install hints

A packaged task source's prerequisite `install` field becomes a union: a plain string (unchanged, meaning "applies to all platforms") or an object keyed by Node's `process.platform` values (`darwin`, `linux`) with a `default` fallback. That one schema widening in `sourceManifestSchema` lets the jira source advertise Linux install commands, and keying by `process.platform` lets crew-config index a hint by platform with no name-mapping table. Motivation: the jira install strings were macOS-only (`brew install ...`), so a Linux user setting the source up saw a Homebrew command that does not apply.

## 2. Editable env defaults for the jira config knobs

The jira manifest's `env` block listed only the three `JIRA_STATE_*` vars, so crew-config pre-filled only those. The four `jira.sh` config knobs (`JIRA_GROUNDCREW_JQL`, `JIRA_REVIEW_PATTERN`, `JIRA_TODO_PATTERN`, `JIRA_DEFAULT_AGENT`) had shell-level defaults but never appeared as editable rows. They now live in `env`. `JIRA_GROUNDCREW_JQL` equals the script's own `:-` default (behavior-neutral, surfaced only for editing); the other three deliberately diverge from the script defaults to seed a sensible starting workflow.

## Requirements

- A prerequisite's `install` may specify different commands per OS (at least macOS and Linux); the jira source provides working Linux commands for `jira` and `jq`, documented in the README.
- The jira manifest's `env` carries all four config knobs alongside `JIRA_STATE_*`, so the TUI pre-fills them as editable rows.
- The added defaults match `jira.sh`'s `${VAR:-default}` fallbacks, with two deliberate exceptions (see Decisions).
- **Backward compatible:** a plain-string `install` still means "all platforms"; every existing string-form manifest validates unchanged.
- **Forward compatible:** the object form is plain nested strings, so an older crew-config that expects a string degrades to showing no hint rather than crashing. No Windows strings.
- The manifest stays the single source of truth for install and env data.

## Decisions

- **Two seeded env defaults deliberately diverge from the `jira.sh` script fallbacks**, to ship a usable workflow without changing unlabeled-ticket dispatch semantics:
  - `JIRA_TODO_PATTERN` -> `New` (script default empty/off): In-Progress-category statuses named "New" demote to `todo` and dispatch as fresh work. The jq guard `$tp != ""` is exactly what lets a non-empty default turn promotion on.
  - `JIRA_REVIEW_PATTERN` -> `In Review` (script default `review`): narrows the in-review promotion to that status name rather than any name containing "review".
  `JIRA_DEFAULT_AGENT` remains empty, matching the script fallback, so Jira issues still need an `agent:` label unless the user opts into a default agent.
  Each is a case-insensitive regex matched against the status name in `jira.sh`'s jq transform, passed via `jq --arg`, so the space in `In Review` is a literal regex character, not a word-split hazard.
- **Injecting the env keys is safe end-to-end.** `manifest.env` reaches list/get/move/verify (via `shellConfigFromManifest` -> `invokeShellCommand`); `JIRA_GROUNDCREW_JQL` and `JIRA_DEFAULT_AGENT` are byte-identical to the script defaults, and crew.config overrides still win over `manifest.env`.
- **One knock-on effect, documented not blocked:** because `manifest.env` shadows `process.env`, a user who `export`s one of these knobs in their shell env (rather than crew.config) now has the shipped default shadow it. This only makes the four new knobs consistent with the three `JIRA_STATE_*` keys, which already behaved this way; the documented config surface (crew.config) is unaffected.
- **The jira `linux` install hint uses the Ubuntu snap** (`sudo snap install jira-cli && sudo snap alias jira-cli jira`) rather than `go install`/release-binary. The snap ships the binary as `jira-cli`; the `snap alias` step keeps `jira` as the single canonical command name on every platform, so the manifest keeps `bin: "jira"` and `jira.sh` invokes `jira` verbatim with no per-OS binary plumbing. The field is a display-only hint, so even if a user skips the alias the prereq probe simply fails and re-surfaces this hint.
- **`JIRA_TOKEN_FILE` intentionally omitted from `env`** - it is a secret path already modeled by the manifest's `secrets` block.
- **Tests colocated in `manifest.test.ts`** (where the shipped `source.json` is already loaded and schema-validated) rather than the spec-named `jiraExampleSource.test.ts`, which is a shell-behavior fixture with no schema/data concern.
- **This is the groundcrew half of a two-repo change.** The crew-config rendering + env pre-fill half shipped separately as crew-config PR #37. Landing this is safe standalone: an older crew-config degrades gracefully.

## Validation

`node --run verify` green (2031 tests, lint, format, spell, markdown, build). New tests cover the union `install` (accepts string, accepts per-OS object, rejects unknown OS key) and the shipped env defaults (`JIRA_DEFAULT_AGENT === "claude"`, `JIRA_TODO_PATTERN === "New"`, `JIRA_REVIEW_PATTERN === "In Review"`). Independent review passes traced the env merge order (behavior-neutrality of `JIRA_GROUNDCREW_JQL`) and confirmed `jira.sh` consumes the patterns via `jq --arg` (space-safe), and confirmed `JIRA_DEFAULT_AGENT` stays empty in the shipped manifest.

Source: plan `2026-07-08-per-os-install-instructions-for-the-jira-task` (plan-keeper `plan-826394705`).


therockstorm:
@mendral-app review

therockstorm:
## Review-body findings

**Already fixed**
- Mendral's prior LGTM review had no actionable findings. Its stale note about `JIRA_DEFAULT_AGENT=claude` is superseded by https://github.com/ClipboardHealth/groundcrew/commit/d1d70df473a7a965f80c718efd9303affaa97f2f, which keeps the shipped Jira default agent empty.

## Conversation-tab comments

**Already fixed**
- `@mendral-app review` was already satisfied by Mendral's review.

```
b4fa3f0620de5f68
c8fe558087f85d9d
```

<sub>🤖 <code>cb-babysit:addressed v1 core@3.15.0</code></sub>

therockstorm (approved):

PR #303, opened 2026-07-14
fix(worktrees): enforce host-dir teardown for scripted worktree cleanup

paulbaranowski:
## What this is

Scripted worktree cleanup now guarantees the on-disk worktree directory is gone when it returns, instead of trusting the remove template to have deleted it.

## Requirements

- A remove template that exits 0 without deleting the directory must not leave the task resurrectable on the next scan.
- Under `--force`, a failing template must not permanently strand a half-provisioned task.
- Without `--force`, a template failure still surfaces as an error (unchanged).
- Abort signals still propagate immediately (unchanged).

## The architecture

Scripted teardown ends with a post-condition check rather than trust: after the remove template runs, `removeScriptedWorktree` verifies the worktree directory is actually gone and scrubs anything left behind via `removeOrphanWorktreeDirectory`, the same disk-cleanup path git-native teardown already uses for orphan recovery.

Previously the function awaited the template and returned, assuming exit 0 meant the directory was deleted. Scripted provisioners can no-op while reporting success (e.g. a remove command paired with `|| true` when the provisioner no longer knows the task), leaving the directory behind; the next `listWorktrees()` scan then finds it and resurrects the task, so cleanup never converges.

Two paths feed the post-condition:

- Template exits 0 but leaves the directory: scrubbed unconditionally.
- Template fails under `--force`: the error is swallowed and cleanup falls back to disk removal, mirroring the git-native orphan recovery in `removeWorktree`. Without `--force` the failure still throws, and a signal abort always rethrows.

Reusing `removeOrphanWorktreeDirectory` means the fallback inherits its existing ENOTEMPTY retry behavior rather than growing a second removal implementation.

## Testing

Four new tests cover the matrix: template exits 0 but leaves the dir, `--force` with a failing template (dir present and dir already gone), and a failing template without `--force` still throwing. Full `node --run verify` (lint, build, coverage-gated tests) is green.

### DIFF
--- a/src/lib/worktrees.test.ts
+++ b/src/lib/worktrees.test.ts
@@ -538,8 +538,8 @@ describe(remove, () => {
     mkdirSync(worktreeDir, { recursive: true });
     userInfoMock.mockReturnValue(makeUserInfo("paul"));
     // Every mocked command returns "" — the template "succeeds" without deleting
-    // the dir. Mirrors a `graft rm ... || true` that no-ops when the provisioner
-    // has no record of the worktree.
+    // the dir. Mirrors a remove command paired with `|| true` that no-ops when
+    // the provisioner has no record of the worktree.
     runCommandMock.mockReturnValue("");
 
     const config = makeConfig({
@@ -548,7 +548,10 @@ describe(remove, () => {
       repositories: [
         {
           name: "billing",
-          provision: { create: "graft new ${branch}", remove: "graft rm ${branch} -f || true" },
+          provision: {
+            create: "provision-tool new ${branch}",
+            remove: "provision-tool rm ${branch} -f || true",
+          },
         },
       ],
     });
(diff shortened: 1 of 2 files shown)


paulbaranowski:
## Review-body findings

### Disagree

- CodeRabbit nitpick (`src/lib/worktrees.ts:606-609`): simplify `options.signal?.aborted === true` to a truthy check. Leaving current behavior: `signal?.aborted === true` is the established convention across this codebase (used at `worktrees.ts:270`, `372`, `525`, `982`, `994` and throughout `src/lib`, e.g. `commandRunner.ts`, `usage.ts`, `util.ts`). Changing this one occurrence would make the file internally inconsistent.

```text
1a5e7b83ac44291f
```


<sub>🤖 <code>pr-babysit:addressed v1 wild-horses@0.2.1</code></sub>

therockstorm (approved):

PR #315, opened 2026-07-21
feat(preLaunch): guard preLaunchEnv against silent empty exports

paulb-instacart:
## What this is

groundcrew now injects a POSIX-shell empty-value guard between `preLaunch` and the launch chain, catching the silent failure where `export VAR="$(cmd)"` masks the substitution's non-zero exit status. Before this, a broken `preLaunch` (missing token file, failed `curl`, unset upstream env) left `VAR` empty, the launch proceeded, and the agent surfaced a downstream 401 with no trail back to preLaunch.

## Requirements

- Fire on the true failure mode — `export VAR="$(cmd)"` returning 0 with `VAR` empty. `set -e` and exit-code checks are already known to miss this.
- Non-aborting — preserve the existing launch-chain contract (prompt read, prompt-dir cleanup, exec must still run).
- POSIX-portable — the emitted shell must run under `/bin/sh` on macOS and `dash` on Linux CI.
- Named diagnostic — an operator reading launch stderr must see the offending variable's name.
- No new configuration surface.

Non-goals: aborting on non-zero `preLaunch` exit (dropped — doesn't catch the motivating bug); scrubbing inherited env on the runner=none path (would silently change runner=none semantics).

## The architecture

The launch chain now carries a per-forwarded-name empty-value probe right after `preLaunch` runs, so a broken export shows up as a stderr line before the launch chain proceeds. A new `preLaunchEmptyCheck` module owns the shell-snippet builder; `launchCommand.ts` splices it into both preLaunch composition sites; `doctor.ts` reflects the runtime gate back to the operator at config time.

Before:

```
unset preLaunchEnv vars → preLaunch → read prompt → exec agent
```

After:

```
unset preLaunchEnv vars → preLaunch → empty-check per name → read prompt → exec agent
```

Each name emits `if [ -z "${NAME-}" ]; then echo "preLaunchEnv: NAME is empty after preLaunch (value length 0)" >&2; fi`. `if…fi` returns 0, so the `&&` chain never short-circuits.

Load-bearing decisions:

- **Warn, don't abort.** Shipping an abort mode next to the warn suggests the operator is protected by two independent layers — they aren't, because the abort layer (non-zero exit) is exactly the case export-masking defeats. Warn-only is the honest scope.
- **Gate the runtime check on `preLaunch !== undefined`.** Listing `preLaunchEnv` names without a `preLaunch` is legal (per `validatePreLaunchEnv`) and means "forward what the parent shell already exported" — nothing this PR can guard. The doctor advisory calls that out explicitly rather than lying.
- **Emit at two splice sites, tolerate asymmetry.** The safehouse chain unsets forwarded names before running `preLaunch`, so the check reads what `preLaunch` produced. The runner=none chain does not unset — a parent-shell value can survive a broken export and hide the failure. Adding a scrub there would silently change runner=none semantics; the caveat is documented at the module and splice sites instead.

What deliberately did not change: launch-chain composition order (the guard slots between existing steps, doesn't replace any), sdx-runner rejection of `preLaunch` (still thrown upstream), `validatePreLaunchEnv` semantics. Safety here rides on the existing `launchCommand.test.ts` and `config.test.ts` suites staying green.

## Interface changes

`crew doctor` gains one `[ok]` advisory per agent with a non-empty `preLaunchEnv`:

After:

```
[ok] preLaunchEnv: claude — forwards 2 var(s): JIRA_API_TOKEN, GITHUB_TOKEN. empty values are warned at launch.
```

When `preLaunch` is unset, the second half switches to `no preLaunch is set, so empty values are NOT checked at launch`, so the copy tracks the runtime gate.

Launch stderr gains one line per empty forwarded var:

```
preLaunchEnv: JIRA_API_TOKEN is empty after preLaunch (value length 0)
```

## Testing

`preLaunchEmptyCheck.test.ts` roundtrips the emitted snippet through a hermetic `/bin/sh` to prove POSIX portability and correct empty/non-empty behavior. `launchCommand.test.ts` adds one composition test per splice site plus an end-to-end runner=none test that exercises the motivating `export VAR="$(cat missing)"` case through a real subshell and asserts the named diagnostic reaches stderr. `doctor.test.ts` covers both advisory copy branches. `node --run verify` passes.

## Follow-ups

Opt-in abort mode behind a config flag, scrubbing inherited env on runner=none to close the asymmetry, and threading the resolved runner into `doctor` so the advisory can suppress itself under sdx are all deferred.

### DIFF
--- a/src/commands/doctor.test.ts
+++ b/src/commands/doctor.test.ts
@@ -940,4 +940,33 @@ describe(doctor, () => {
     expect(lines).toMatch(/requested=cmux/);
     expect(lines).toContain("cmux binary is not on PATH");
   });
+
+  it("emits an advisory line for each agent with a non-empty preLaunchEnv", async () => {
+    loadConfigMock.mockResolvedValue(
+      makeConfig({
+        default: "claude",
+        definitions: {
+          claude: {
+            cmd: "claude",
+            color: "#fff",
+            preLaunchEnv: ["JIRA_API_TOKEN", "GITHUB_TOKEN"],
+          },
+          codex: { cmd: "codex", color: "#3267e3" },
+        },
+      }),
+    );
+
+    const actual = await doctor();
+
+    expect(actual).toBe(true);
+    const lines = consoleLog.output();
+    // Advisory renders with the `[ok]` tag (ok=true, required=false) — same
+    // pattern as `local runner (none)`. The informational nature is carried by
+    // the wording (`forwards N var(s): …`), not by a fail tag.
+    expect(lines).toMatch(/\[ok]\s*preLaunchEnv: claude/);
+    expect(lines).toContain("JIRA_API_TOKEN");
+    expect(lines).toContain("GITHUB_TOKEN");
+    // Agent without preLaunchEnv must not appear as an advisory line.
+    expect(lines).not.toMatch(/preLaunchEnv: codex/);
+  });
 });
--- a/src/commands/doctor.test.ts
+++ b/src/commands/doctor.test.ts
@@ -949,6 +949,7 @@ describe(doctor, () => {
           claude: {
             cmd: "claude",
             color: "#fff",
+            preLaunch: 'export JIRA_API_TOKEN="tok"',
             preLaunchEnv: ["JIRA_API_TOKEN", "GITHUB_TOKEN"],
           },
           codex: { cmd: "codex", color: "#3267e3" },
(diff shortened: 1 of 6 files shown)


paulb-instacart:
## Review-body findings

### Already fixed

- **Mendral / Code Review — LGTM.** No findings to act on. Verified: shell injection prevented by upstream `ENV_VAR_NAME_PATTERN` in `config.ts`; `if [ -z "${NAME-}" ]` is POSIX-correct; `&&`-chain short-circuit semantics preserved.

Recording fingerprint so future babysit passes dedupe.

```
05b5063b31291370
```


<sub>🤖 <code>pr-babysit:addressed v1 wild-horses@0.2.1</code></sub>

paulb-instacart:
@therockstorm could you review?

therockstorm (approved):

PR #317, opened 2026-07-23
feat: link cmux tasks to source issues (TG-4209)

Davidster:
## Summary

- Propagate each task source's exact URL through workspace launch metadata on dispatch and resume.
- Show a compact `Linear ↗` link in cmux sidebar cards, with only the visible label acting as the click target.
- Leave URL-less tasks, workspace identity, cleanup, and non-cmux adapters unchanged.

## Why

Groundcrew already receives and persists the task source's direct URL, but cmux cards did not expose it. This makes the issue reachable from the card without constructing a URL from the task identifier or changing how workspaces are identified and managed.

## Validation

- `node --run verify` — all 9 checks passed; 85 test files and 2,079 tests passed with 100% coverage.
- Automated tests cover cmux metadata commands for tasks with and without a source URL.
- Resume tests verify cached URLs, legacy Linear URL recovery, and that URL-less tasks from other sources cannot inherit a colliding Linear URL.
- Live A/B smoke test with `TG-4273`:
  - the before state had no issue link;
  - the after state rendered `Linear ↗` as a hyperlink;
  - activating the label opened the exact source-provided Linear issue URL;
  - clicking the blank area to the right of the label did not open the link.

### Before

cmux sidebar card before: no Linear issue link

### After

cmux sidebar card after: compact Linear issue link

## Notes

- cmux does not expose link metadata on its workspace-title row, so the issue link is a dedicated sidebar status row. This design was confirmed in the live smoke test.
- The exact source-provided URL is forwarded; no Linear URL is synthesized from the task identifier.
- This change and the Safehouse hook fix in #334 are independent and may merge in either order.

Closes tg-4209

Agent session: `019f9081-657e-7cb3-9f26-f4ef1263e855`

<sub>🤖 <code>cb-ship:updated v1</code></sub>

### DIFF
--- a/src/commands/resumeWorkspace.ts
+++ b/src/commands/resumeWorkspace.ts
@@ -36,6 +36,7 @@ export interface ResumeWorkspaceOptions {
 interface TaskDetails {
   title: string;
   description: string;
+  url?: string;
 }
 
 interface ResumeContext {
@@ -45,6 +46,7 @@ interface ResumeContext {
   worktree: WorktreeEntry;
   title: string;
   description: string;
+  url?: string;
   completionTaskId: string;
   completionMarkDoneSupported: boolean;
   reason?: string;
@@ -74,6 +76,7 @@ async function fetchTaskDetails(task: string): Promise<TaskDetails | undefined> 
     return {
       title: issue.title,
       description: issue.description ?? "",
+      url: issue.url,
     };
   } catch (error) {
     log(`Resume Linear detail lookup failed for ${task}: ${errorMessage(error)}`);
@@ -95,6 +98,7 @@ async function contextFromLinear(
     worktree,
     title: resolved.title,
     description: resolved.description,
+    url: resolved.url,
     completionTaskId,
     completionMarkDoneSupported: taskSupportsCompletionCommand({
       rawSources: sourcesFromConfig(config),
@@ -115,6 +119,7 @@ async function contextFromState(
   // enrich the prompt title/description (which falls back to the task id).
   const details = isLinearEnabled(config) ? await fetchTaskDetails(task) : undefined;
   const completionTaskId = state.completionTaskId ?? task;
+  const url = state.url ?? details?.url;
   return {
     task,
     repository: state.repository,
@@ -125,6 +130,7 @@ async function contextFromState(
     worktree: { ...worktree, branchName: state.branchName },
     title: details?.title ?? state.title ?? task.toUpperCase(),
     description: details?.description ?? "",
+    ...(url === undefined ? {} : { url }),
     completionTaskId,
     completionMarkDoneSupported: taskSupportsCompletionCommand({
       rawSources: sourcesFromConfig(config),
@@ -274,6 +280,7 @@ export async function resumeWorkspace(
       config,
       name: task,
       displayName: context.title,
+      ...(context.url === undefined ? {} : { url: context.url }),
       cwd: launchDir,
       command: launchCmd,
       agent: context.agent,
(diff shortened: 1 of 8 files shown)


src/lib/cmuxAdapter.ts
```diff
@@ -44,17 +44,24 @@ export const cmuxAdapter: Adapter = {
       );
       throw new Error(`Unexpected cmux output: ${output}`);
     }
+    if (spec.url !== undefined) {
+      await applyCmuxStatusBestEffort({
+        workspaceId,
+        key: "task",
```
Davidster, line 50: what does key: "task" do? that's new?

src/commands/resumeWorkspace.ts
```diff
@@ -135,6 +143,25 @@ async function contextFromState(
   };
 }
 
+function taskUsesLinearSource(arguments_: { config: ResolvedConfig; taskId: string }): boolean {
+  const { config, taskId } = arguments_;
+  const rawSources = sourcesFromConfig(config);
+  const colonIndex = taskId.indexOf(":");
```
Davidster, line 149: is checking for the colon robust? when does the string have and not have a colon?

therockstorm (approved):

src/commands/resumeWorkspace.ts
ClipboardJason: if I recall this was my doing. It now stores the definitive task ID in the description of the CMUX pane rather than the title because there is an option to set the CMUX pane title to the ticket title rather than the ticket ID. So the source of truth for the ID is now hidden in the summary and needs to be parsed. Since they're only ever constructed by the same scripts I think parsing it like this is good enough for now.

src/lib/cmuxAdapter.ts
Davidster: Leaving current behavior. `task` is cmux’s status-slot key: it creates a task-link row distinct from the existing `agent` row, so each can be updated independently. This key is introduced by this PR.

<sub>🤖 <code>cb-babysit:addressed v1 skill@1.0.3</code></sub>

src/commands/resumeWorkspace.ts
Davidster: Leaving current behavior. New run state stores canonical IDs as `<source-name>:<natural-id>`; the no-colon form only exists in legacy state. For that legacy case, this code infers Linear only when exactly one source is configured, and refuses to infer when multiple sources are possible.

<sub>🤖 <code>cb-babysit:addressed v1 skill@1.0.3</code></sub>

src/lib/cmuxAdapter.ts
```diff
@@ -201,20 +208,54 @@ function extractCmuxOpenId(output: string): string | undefined {
...
+}
+
+function cmuxTaskLinkText(url: string): string {
+  let label: string;
+  try {
+    label = new URL(url).hostname === "linear.app" ? "Linear ↗" : "Issue ↗";
+  } catch {
+    label = "Issue ↗";
+  }
+  return `[${label}](${url})`;
```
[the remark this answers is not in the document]
Davidster: Addressed in https://github.com/ClipboardHealth/groundcrew/commit/cc6fde52efa7bdbdcf6720782d2dd8292b42e083. Markdown destinations now escape backslashes and closing parentheses and percent-encode line breaks, with public workspace-open coverage for each case.

<sub>🤖 <code>cb-babysit:addressed v1 skill@1.0.3</code></sub>

Davidster:
## Review-body findings

**Agree**

- CodeRabbit’s Markdown-destination injection finding was valid and is addressed in https://github.com/ClipboardHealth/groundcrew/commit/cc6fde52efa7bdbdcf6720782d2dd8292b42e083.

**Already fixed**

- Mendral reported LGTM with no actionable findings; no additional change was required.

```text
9da531b8182131ca
56f295510987cfed
```

<sub>🤖 <code>cb-babysit:addressed v1 skill@1.0.3</code></sub>

PR #318, opened 2026-07-23
fix: keep cmux Codex hooks valid under Safehouse

Davidster:
## Summary

- keep cmux-generated Codex hook output valid when the hook invokes cmux inside Safehouse
- redirect Foundation cache access to the already-staged writable Codex home and disable cmux CLI telemetry for those hook invocations
- leave user hooks and non-command hooks unchanged, and make the rewrite idempotent

## Why

#311 made sandboxed Codex load cmux hooks from Groundcrew’s relocated `CODEX_HOME`. Each lifecycle event then began launching the cmux CLI inside Safehouse. The Sentry Cocoa launch profiler initializes before cmux `main`, tries to access `~/Library/Caches/io.sentry`, and writes its cache failure diagnostics to stdout. That turns cmux’s valid `{}` hook response into invalid Codex hook JSON.

`CMUX_CLI_SENTRY_DISABLED` is cmux’s post-start telemetry switch, but it cannot prevent the pre-`main` launch-profiler cache lookup. `CFFIXED_USER_HOME` redirects that Foundation lookup into the writable staged home without granting the sandbox access to the user’s real cache directory. Both exports are scoped to commands generated by cmux.

The root fix belongs upstream in cmux: JSON hook commands should not initialize launch profiling or emit diagnostics on stdout. This PR is the compatibility fix for released cmux versions.

## Testing

- `node --run verify`
- 84 test files, 2,043 tests, 100% coverage
- manually reproduced the failure as `{}` followed by Sentry diagnostics and confirmed the hardened hook emits exactly `{}`

### DIFF
--- a/src/lib/cmuxAgentHookInstall.test.ts
+++ b/src/lib/cmuxAgentHookInstall.test.ts
@@ -1,3 +1,7 @@
+import { mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import os from "node:os";
+import path from "node:path";
+
 import { installCmuxAgentHooks } from "./cmuxAgentHookInstall.ts";
 
 const runCommandMock = vi.hoisted(() =>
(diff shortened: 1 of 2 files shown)


Davidster:
Closing this draft because a fresh end-to-end test using the current Groundcrew setup (Safehouse + cmux) completed without the invalid Stop-hook JSON failure, and we found no Groundcrew-installed Stop hook in that session. Since the failure is no longer reproducible, the CFFIXED_USER_HOME / Sentry environment workaround is not justified by the evidence. We can reopen or replace this with a narrower fix if the issue recurs with a reproducible case.

PR #334, opened 2026-08-03
fix: forward cmux Safehouse writable grants (TG-4209)

Davidster:
## Why

cmux 0.64.21+ runs Codex lifecycle hooks from `~/.cmux/hooks`, and its CLI initializes Sentry while producing the JSON those hooks return. Inside Groundcrew's Safehouse sandbox, cmux could not read the hook executables or write `~/Library/Caches/io.sentry`. That caused every lifecycle hook to fail, first with exit code 126 and then with invalid JSON when Sentry diagnostics contaminated stdout.

[core-utils #831](https://github.com/ClipboardHealth/core-utils/pull/831) added the required cmux paths to Clearance 1.6.6. Groundcrew must consume the new writable-path field for that fix to reach launched agents. This has no direct customer impact; it restores reliable cmux sidebar activity reporting for Groundcrew operators using Codex with Safehouse.

## Summary

- Upgrade `@clipboard-health/clearance` from 1.6.3 to 1.6.6.
- Forward Clearance's writable cmux paths into the agent Safehouse wrapper.
- Merge those paths with Groundcrew's existing staged `CODEX_HOME` grant.
- Cover the generated Safehouse command with a focused regression test.

## Validation

- Red: `npm exec vitest -- run src/lib/agentLaunch.test.ts` failed because `/Users/dev/Library/Caches/io.sentry` was absent from `--add-dirs`.
- Green: `npm exec vitest -- run src/lib/agentLaunch.test.ts src/lib/launchCommand.test.ts` — 108 tests passed.
- `node --run verify` — passed.
- Pre-push verification — passed.

Before the fix, a real Groundcrew-launched Codex session reported:

```text
SessionStart hook (failed)
error: hook exited with code 126

PreToolUse hook (failed)
error: hook returned invalid pre-tool-use JSON output

PostToolUse hook (failed)
error: hook returned invalid post-tool-use JSON output

Stop hook (failed)
error: hook exited with code 126
```

The minimized failing hook command showed why the JSON was invalid: cmux returned `{}` and then wrote Sentry cache errors to stdout, so `jq` exited 5.

```text
{}
[Sentry] [error] Can't create base path ... /Users/davidhuculak/Library/Caches/io.sentry ...
jq: parse error: Invalid numeric literal at line 2, column 8
json_exit=5
```

After applying the Clearance grants through Groundcrew, a fresh cmux/Codex smoke session ran the same read-only prompt without any hook failures:

```text
Ran printf 'hook smoke test\n'
  └ hook smoke test

Output: hook smoke test
No files changed and no PR opened.
```

## Notes

- Depends on merged [core-utils #831](https://github.com/ClipboardHealth/core-utils/pull/831).
- Related Linear issue: [TG-4209](https://linear.app/clipboardhealth/issue/TG-4209/link-cmux-sidebar-task-headings-to-their-linear-issues).
- This compatibility fix is intentionally separate from draft Groundcrew [#317](https://github.com/ClipboardHealth/groundcrew/pull/317). #317 will be rebased after this PR merges.

Agent session: `codex resume 019f9081-657e-7cb3-9f26-f4ef1263e855`

<sub>🤖 <code>cb-ship:created v1 skill@1.0.1</code></sub>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,15 +1,15 @@
 {
   "name": "@clipboard-health/groundcrew",
-  "version": "4.49.2",
+  "version": "4.49.3",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@clipboard-health/groundcrew",
-      "version": "4.49.2",
+      "version": "4.49.3",
       "license": "MIT",
       "dependencies": {
-        "@clipboard-health/clearance": "1.6.3",
+        "@clipboard-health/clearance": "1.6.6",
         "@linear/sdk": "89.0.0",
         "agent-trust": "1.0.0",
         "cosmiconfig": "9.0.2",
@@ -1595,9 +1595,9 @@
       "license": "MIT"
     },
     "node_modules/@clipboard-health/clearance": {
-      "version": "1.6.3",
-      "resolved": "https://registry.npmjs.org/@clipboard-health/clearance/-/clearance-1.6.3.tgz",
-      "integrity": "sha512-/AEXdHQg4omWSxQxpjISXM3Nv4t4Bm4uSioCTX6N47iv5sBTTFldVcRvy+BRX4sqkjWbpRMYYoER6rFA6VvobA==",
+      "version": "1.6.6",
+      "resolved": "https://registry.npmjs.org/@clipboard-health/clearance/-/clearance-1.6.6.tgz",
+      "integrity": "sha512-ivIlc9AseWX2NhBCHtCZle7lPFqoGT6u/NIdWZ1zAvknVvztGk4gHmeoJ0LD4woabsH85CxVee4gAUwT/wnOGQ==",
       "license": "MIT",
       "dependencies": {
         "tslib": "2.8.1"
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "verify": "node scripts/verifyAll.mts"
   },
   "dependencies": {
-    "@clipboard-health/clearance": "1.6.3",
+    "@clipboard-health/clearance": "1.6.6",
     "@linear/sdk": "89.0.0",
     "agent-trust": "1.0.0",
     "cosmiconfig": "9.0.2",
--- a/src/lib/agentLaunch.test.ts
+++ b/src/lib/agentLaunch.test.ts
@@ -183,6 +183,20 @@ describe(composeAgentLaunch, () => {
     expect(launchCommand).toContain('"$@"');
   });
 
+  it("adds cmux writable Safehouse grants supplied by clearance", () => {
+    resolveSafehouseCmuxIntegrationMock.mockReturnValue(
+      safehouseCmuxIntegrationFixture({
+        addDirs: ["/Users/dev/Library/Caches/io.sentry"],
+      }),
+    );
+
+    const launchCommand = compose();
+
+    expect(launchCommand).toContain(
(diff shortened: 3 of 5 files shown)


therockstorm (approved):

ClipboardJason:
I chased the same 126 from the cmux side yesterday and landed on the same root cause, so this matches what I found. One gap worth flagging before it merges: the grants fix the *errors*, but I don't think they restore the sidebar reporting the description claims, and the failure mode is silent.

**The hooks still can't reach cmux.** Every cmux hook talks to the daemon over the unix socket, and safehouse denies AF_UNIX connects by default (`20-network.sb` allows `(remote ip)` plus one literal for mDNSResponder). I unpacked published `@clipboard-health/clearance@1.6.6` and `clearance-only.sb` is unchanged from 1.6.3, still only remote-IP shaping. Neither this PR nor core-utils #831 adds a rule.

The grants can't cover it, because `connect()` classifies as `network-outbound` rather than file access. Both forms fail:

```
--add-dirs=$CMUX_SOCKET_PATH             → Operation not permitted, errno 1
--add-dirs-ro=$(dirname $CMUX_SOCKET_PATH) → Operation not permitted, errno 1   ← what resolveCmuxReadOnlyDirs grants
```

**Measured with the post-PR grant set.** I ran a real `codex exec` under safehouse with exactly the grants this PR delivers (`~/.cmux/hooks` ro, `~/Library/Caches/io.sentry` rw, `dirname(socket)` ro, `SAFEHOUSE_CMUX_ENV_PASS`). No 126 and no invalid JSON, so those two failure modes are genuinely gone. But no cmux hooks fired at all: the wrapper pings the socket, the ping fails, and it passes through without injecting. Only a plugin `SessionStart` showed up.

**Why the smoke test reads green either way.** Both hook shapes fail open. In-sandbox, with the socket denied:

```
cmux --socket "$CMUX_SOCKET_PATH" ping        → Operation not permitted   (exit 1)
cmux hooks feed --source codex --event ...    → {}                        (exit 0)
```

The Pre/PostToolUse scripts background the real call and `echo '{}'` unconditionally. "Ran the same read-only prompt without any hook failures" is equally consistent with working and with delivering nothing, so it can't separate the two. Two checks that can: run `cmux --socket "$CMUX_SOCKET_PATH" ping` inside the wrap, or confirm the session actually shows up in the sidebar.

**What closes it.** One rule, appended to the profile:

```lisp
(allow network-outbound (literal "/private/tmp/cmux-<build>.sock"))
```

Worth noting the resolved path: `/tmp` is a symlink to `/private/tmp`, and seatbelt matches post-resolution, so a `/tmp/...` literal silently fails to match. The socket basename is also build-specific (installed vs. Xcode dev builds differ), so it wants generating from `CMUX_SOCKET_PATH` rather than hardcoding. With that rule plus this PR's grants I get `PONG` in-sandbox and a full `codex exec` with every hook injected and completed.

Delivery is the awkward part: `--append-profile` has no env var, and the `<workdir>/.safehouse` route is out because trust is exact-dir match against `~/.config/safehouse/trusted-workdirs`, so per-task worktrees can never be pre-trusted. That leaves adding the rule to `clearance-only.sb` (it already owns an `--append-profile`, so no new plumbing), or a groundcrew `local.safehouse.appendProfiles` key mirroring the existing `enable`. Safehouse's own `docker` feature is the precedent for a per-feature socket literal, so a `cmux` feature upstream would be the tidiest long-term home.

None of this blocks merging as a strict improvement over the current state. It's the description I'd narrow: this stops the hooks erroring, but activity reporting stays dark until the socket opens.

PR #338, opened 2026-08-05
fix(cleaner): stop re-logging identical cleanup failures every poll

ClipboardJason:
## Why

A single worktree the orchestrator can never tear down produces one identical log line per poll, forever. On my machine that is 18,138 lines / 19.4 MB of:

```
[15:48:57] Cleanup failed for tg-4259 (host): worktree has 1 modified file and 2 untracked files. Run crew cleanup --force tg-4259 to discard them, or commit/stash in /Users/jason/.../cbh-admin-frontend-tg-4259 first.
```

`createCleaner` runs `reapWorktrees` on every 120s poll. Teardown hits the dirty-worktree data-loss guard, reports the same failure, and `logTeardown` prints it verbatim — with no memory of having just printed it. The `event=cleanup outcome=failed` line doubles the volume. The operator's main pane, which `zellijAdapter` tails from the log file, becomes unreadable.

This PR fixes the reporting. It deliberately does **not** touch the dirty-worktree guard or the `npm ci` hook that dirties worktrees in the first place — those are separate PRs.

## What

A new `src/lib/repeatedFailures.ts` holds a per-orchestrator, in-memory recurrence log. `createRepeatedFailureLog().observeAll({ occurrences })` takes one poll's complete set of failures and classifies each:

| recurrence   | when                        | reported as                                                   |
| ------------ | --------------------------- | ------------------------------------------------------------- |
| `first`      | key not seen last poll      | the existing line, verbatim                                     |
| `changed`    | same key, different message | the existing line, verbatim                                     |
| `ongoing`    | every 30th unchanged repeat | `Cleanup still blocked for <task> (<kind>) after N attempts:`    |
| `suppressed` | any other unchanged repeat  | nothing                                                         |

`createCleaner` constructs one log and threads it into `reapWorktrees` through a new options object (`{ signal?, failureLog? }`). One-shot callers — `crew cleanup` via `logTeardown`, and the reviewer's single-entry fast path — pass no log and keep reporting every occurrence verbatim, so a user who just asked for a cleanup is never shown a blank where a reason should be.

For one stuck worktree that is roughly 1,440 lines/day down to roughly 48.

### Design notes

- **Teardown still runs every poll.** Only the reporting is coalesced. The attempt is one `git status --porcelain`; skipping it would delay real cleanup once the operator clears the dirt. This is not backoff.
- **The first occurrence is never swallowed.** The operator learns about a blocker on the poll it appears.
- **A changed message always re-announces.** A change is real news, and a silent slot would otherwise let a different failure (worktree locked, disk full) hide behind a suppressed one.
- **Not log-once-and-never-again.** An operator attaching an hour later would see a healthy-looking log. The hourly "still blocked after N attempts" heartbeat is the deliberate cost of not going fully silent.
- **`observeAll` forgets keys absent from the batch**, so a failure that clears and later recurs is announced again. Documented on the interface: callers must hand over every failure they track, not a subset.
- **In-memory and per-`orchestrate()`.** A restarted orchestrator re-announces every still-blocked worktree once — deliberate; the person who just restarted is exactly the person who wants to see current blockers.

### Telemetry change worth a reviewer's attention

`event=cleanup outcome=failed` becomes edge-triggered rather than one-per-poll, and carries a new cumulative `repeats=N`. `max(repeats)` recovers the true occurrence count exactly. There is no machine consumer of this stream today (`grep -rn 'getLogFile' src` finds only `zellijAdapter` tailing the file for a human), and leaving it per-poll would have left roughly half the spam in the file the operator actually watches.

The third commit demotes `Cleaning up N terminal worktree(s)` from `log()` to `debug()` — it repeats every poll for the same reason and is fully subsumed by the per-entry lines that follow. Kept as its own commit so it can be dropped independently.

## Validation

Red/green/refactor throughout: `src/lib/repeatedFailures.test.ts`, the new cases in `src/commands/teardownReporter.test.ts`, and the user-visible regression in `src/commands/cleaner.test.ts` were each written and watched fail before the implementation existed.

The regression test that encodes the bug:

```
logs an unresolvable cleanup failure once across polls
  -> one "Cleanup failed for team-1" line after 3 polls
  -> teardown still called 3 times
```

`node --run verify`:

- `test`: 13 failed | 2096 passed (2109), 86 files. All 13 are the pre-existing `openWorkspace.test.ts` / `cmuxIntegration.addDirs is not iterable` failures present on `main`. No new failures.
- Coverage of the three touched source files from their own suites: 100% statements / branches / functions / lines (93/93, 48/48, 15/15, 92/92).
- `architecture:check`, `cpd`, `format:check`, `knip`, `markdown:lint`, `syncpack:lint`: green.
- `build`: only the pre-existing `agentLaunch.ts(164,26) TS2339 addDirs` error.
- `lint`: fails locally with `Rule 'no-top-level-await' not found in plugin 'node'` — my sandbox's `node_modules` carries oxlint 1.73.0 while `main` pins 1.76.0. Re-run under `oxlint@1.76.0` the branch is clean apart from the same pre-existing `addDirs` errors. CI should be green here.

### DIFF
--- a/src/commands/cleaner.test.ts
+++ b/src/commands/cleaner.test.ts
@@ -6,7 +6,7 @@ import type { BoardState } from "../lib/taskSource.ts";
 import { type WorktreeEntry, worktrees } from "../lib/worktrees.ts";
 import { captureConsoleLog, type ConsoleCapture } from "../testHelpers/consoleCapture.ts";
 import { emptyTeardownResult } from "../testHelpers/teardownResult.ts";
-import { createCleaner } from "./cleaner.ts";
+import { type Cleaner, createCleaner } from "./cleaner.ts";
 
 vi.mock(import("../lib/worktrees.ts"), async (importOriginal) => {
   const actual = await importOriginal();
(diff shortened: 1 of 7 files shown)


ClipboardJason:
Closing — we've decided not to pursue this. The per-poll log volume is being addressed at the source by #339 (hook-generated files no longer block teardown, so these failures stop recurring rather than being deduplicated after the fact). Branch left in place if we want to revisit.

PR #379, opened 2026-08-16
feat(usage): record agent token usage against the task

lgoyal6:
## Why

`status` already reports wall clock and pull request outcomes, so tokens were the
only thing missing before cost per task became computable. It turns out they cost
nothing to collect: Claude writes a JSONL transcript per session and stamps every
assistant message with its own `usage`, so there is no API call, no provider
lookup, and no second run involved. `usage.ts` currently tracks quota percentages
from codexbar, which tell you how close a run is to a limit but never what it
cost.

That gap is why `agent-any` routes on session headroom alone. Once the tokens are
recorded, the routing decision has a real input rather than a proxy for one.

## What

The `SessionEnd` hook groundcrew already installs now also invokes
`groundcrew record-usage`, which reads the transcript path from the hook payload
on stdin, sums the session, and folds the totals into the task's run state as
`RunState.usage`.

The hook is guarded on `GROUNDCREW_TASK_ID` and swallows every failure, the same
shape as the `set-progress` hook beside it, because a hook should never be able to
break the agent it is observing. A missing count degrades reporting rather than
failing a run.

## Two decisions worth reviewing

**Usage is counted once per message id.** A transcript contains the same assistant
message repeatedly, because it is rewritten while streaming and again when a
session resumes. On a real transcript, 718 records carried usage but only 327
message ids were distinct, and summing every record inflated the total **2.13x**.
The inflation is not uniform across fields either, 2.10x on cache reads against
2.64x on output, so it cannot be corrected with a constant after the fact. It
distorts the input/output mix as well as the total.

**The four fields stay separate.** Cache reads bill at roughly a tenth of fresh
input, so a single `totalTokens` would price a cache-heavy session as though none
of it had been cached. This module counts and deliberately does not price; a rate
sheet belongs wherever pricing is decided.

## Scope and limits

Claude only. Codex installs hooks through a different mechanism and I could not
verify that it exposes an equivalent transcript, so rather than report a confident
zero for codex runs, `usage` is simply absent there. **Consumers must read a
missing `usage` as unknown, never as zero.** Wiring codex up is a follow-up for
someone who knows that CLI better than I do.

Nothing consumes the field yet. Surfacing it in `status` output is deliberately
left out to keep this reviewable as a single concept.

## Validation

`node --run verify` passes, including the 100% coverage thresholds.

- 20 tests on the parser, covering partially written transcripts, records without
  ids, non-object JSON lines, and the deduplication above
- 13 tests on the command, covering the stdin payload path, a missing transcript,
  a session that outlives its run state, and lifecycle preservation
- 4 tests on the hook wiring
- Full suite: 2,209 tests passing, coverage 100% on statements, branches,
  functions and lines

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -6,6 +6,7 @@ import { doctor } from "./commands/doctor.ts";
 import { initConfigCli } from "./commands/init.ts";
 import { interruptWorkspaceCli } from "./commands/interruptWorkspace.ts";
 import { openWorkspaceCli } from "./commands/openWorkspace.ts";
+import { recordUsageCli } from "./commands/recordUsage.ts";
 import { orchestrate } from "./commands/orchestrator.ts";
 import { resumeWorkspaceCli } from "./commands/resumeWorkspace.ts";
 import { setupWorkspaceCli } from "./commands/setupWorkspace.ts";
@@ -187,6 +188,12 @@ const SUBCOMMANDS: Record<string, Subcommand> = {
     summary: "List, get, create, and complete tasks across configured sources",
     usage: "<list|get|create|done|validate> [...]",
     invoke: taskCli,
+  },
+  "record-usage": {
+    summary: "Record an agent session's token usage against a task",
+    usage: "--task <task> [--transcript <path>]",
+    hidden: true,
+    invoke: recordUsageCli,
   },
   status: {
     summary: "Print read-only groundcrew state, or one task's local/Linear status",
--- a/src/commands/recordUsage.test.ts
+++ b/src/commands/recordUsage.test.ts
@@ -1,0 +1,191 @@
+import { readFileSync } from "node:fs";
+
+import { loadConfig, type ResolvedConfig } from "../lib/config.ts";
+import { readRunState, type RunState, updateRunState } from "../lib/runState.ts";
+import { recordUsageCli } from "./recordUsage.ts";
+
+vi.mock("node:fs", async (importOriginal) => {
+  const actual = await importOriginal<typeof import("node:fs")>();
+  return { ...actual, readFileSync: vi.fn<typeof readFileSync>() };
+});
+vi.mock(import("../lib/config.ts"), async (importOriginal) => {
+  const actual = await importOriginal();
+  return { ...actual, loadConfig: vi.fn<typeof loadConfig>() };
+});
+vi.mock(import("../lib/runState.ts"), async (importOriginal) => {
+  const actual = await importOriginal();
+  return {
+    ...actual,
+    readRunState: vi.fn<typeof readRunState>(),
+    updateRunState: vi.fn<typeof updateRunState>(),
+  };
+});
+
+const readFileSyncMock = vi.mocked(readFileSync);
+const loadConfigMock = vi.mocked(loadConfig);
+const readRunStateMock = vi.mocked(readRunState);
+const updateRunStateMock = vi.mocked(updateRunState);
+
+const config = { logging: { file: "/tmp/groundcrew.log" } } as ResolvedConfig;
+
+function runState(overrides: Partial<RunState> = {}): RunState {
+  return {
+    task: "ENG-1",
+    repository: "repo-a",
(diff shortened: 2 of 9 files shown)


lgoyal6:
Good catch, and it was worse than reported. Fixed in b3fdb8a.

**The reported bug.** `parseRunState` builds `RunState` field by field and did not read `usage`, so counts were written to disk and dropped on the way back. A resumed task folded into an empty total rather than accumulating, and any later lifecycle write spread the parsed record and erased the field from disk. My test missed it because it mocked `readRunState`, so it never exercised the parser at all. That was a fair criticism of the test, not just the code.

**A second instance, not in the review.** `recordRunState` has the same bug for a different reason. It reconstructs the record rather than spreading it, carrying `title`, `url`, `completionTaskId` and `adoptedBranch` forward by hand. Token counts were not on that list, so every resume through that path reset a task's spend to zero. Fixed alongside.

**Where the parser lives.** I put `parseTokenUsage` in `tokenUsage.ts` rather than inline in `parseRunState`, so the field names exist in exactly one place. A reader that drifts from the writer loses counts silently, which is the failure this whole thing is guarding against. A malformed total on disk now reads as zeroes rather than rejecting the record: a field that fails to parse should cost you that field, not the task's entire state.

**Tests now use the real store.** Five cases in `runState.test.ts` covering the round trip, a lifecycle transition that never mentions usage, a resume that rebuilds the record, absent staying distinguishable from zero, and a malformed total. Four of them fail without this commit, so they are regression coverage rather than tests written to match the code.

**One thing to flag.** `parseRunState` and `recordRunState` were both already at the complexity ceiling, so adding a field pushed them to 21 and 22 against a max of 20. Rather than suppress the rule I extracted the optional-field assembly into `optionalFields` and `carriedFields`. That is a non-functional change riding along with a fix, which your `gitWorkflow.md` would normally have me split out. It is here because the lint gate blocks the fix otherwise, but I am happy to pull it into its own PR if you would prefer that.

`node --run verify` passes: 2,236 tests, 100% on statements, branches, functions and lines.

lgoyal6:
On the first-wins point in `tokenUsage.ts`: acted on it, but the premise does not reproduce here, so I want to show the evidence rather than just say "fixed".

**What I measured.** Across the transcripts I can read, 271 message ids carry more than one usage record, and **zero of them have counts that differ between records**. First-wins and last-wins produce byte-identical totals, 192,618,403 on the sample I checked. The duplicates in this format are exact repeats, not growing streaming snapshots.

**Why I changed it anyway.** The shape is not guaranteed across agent versions, and the failure is silent in both directions. The previous commit fixed a 2.13x overcount; getting this wrong the other way would have produced an undercount just as invisible. Not worth leaving to chance for a few lines.

**What it does now.** Counts collapse to the **highest value seen per field per message id**, in d4ce8c8. That is correct under both shapes: where repeats are identical every candidate is equal and the maximum is that value, so this is a no-op on current data and the total is unchanged at 192,618,403; where they are cumulative, the maximum is the final complete count.

I went with the maximum rather than last-wins deliberately. Last-wins is correct only if the trailing record is always the complete one, and it silently undercounts if a partial ever lands last. The maximum has no such failure mode. Fields are compared independently, so a repeat that grows on output while reporting less input cannot drag the input count down.

Three regression tests cover a growing repeat, a shrinking one, and per-field comparison.

`node --run verify` passes: 2,239 tests, 100% on statements, branches, functions and lines.

therockstorm:
Thanks for putting this together. The underlying usage-accounting problem is worth solving, but after reviewing the implementation I don’t think we should merge this approach. Usage collection needs to be host-owned, idempotent across resumed sessions, and independent of the workspace backend and sandbox runner. Reworking this PR toward that design would be a substantial architectural change rather than a normal review iteration. I also don’t have the capacity right now to shepherd that redesign responsibly, so I’m going to close this rather than leave you with an open-ended request for changes. I appreciate the investigation and the transcript-deduplication work—it helped clarify what a future implementation needs to handle.

PR #388, opened 2026-08-28
feat(safehouse): auto-resolve symlinked agent config into read-only grants

bionboy:
## Problem

When `~/.claude` is managed by a dotfiles repo (stow, chezmoi, yadm, dotbot), every entry under it is a symlink into somewhere like `~/.dot_files/config/claude/`. Safehouse grants `~/.claude`, but macOS seatbelt resolves symlinks *before* evaluating policy, so the kernel checks the ungranted target path. The agent gets `EPERM: operation not permitted, stat '/Users/<user>/.claude/settings.json'` and Claude Code silently skips the whole settings file — plus skills, hooks, agents, commands, and CLAUDE.md.

The error is misleading: it names a file that plainly exists and is readable outside the sandbox.

## Change

New `src/lib/sandboxSymlinkGrants.ts` resolves symlinked config paths at launch and appends the resolved targets to the existing `--add-dirs-ro` list. Gated to `runner: "safehouse"` — `sdx` and `none` are untouched. Merged with `local.readOnlyDirs` (explicit entries first), deduped, then run through the existing `existsSync` filter.

### Scope: agent config dir + a tool allowlist

Resolution covers `~/.claude` plus `~/.config/gh`, `~/.config/git`, and `~/.gitconfig`.

- **agent-config-only is too narrow**: it leaves `gh` unable to start under a dotfiles-managed `config.yml`, so an agent cannot open the PR it was dispatched to open. This is not hypothetical — it blocked the first dispatch of this ticket.
- **all of `~/.config` is too wide**: a typical dotfiles setup symlinks a dozen unrelated entries (`nvim`, `ranger`, `raycast`, `opencode`, …), and following all of them is a large silent inference.
- The allowlist covers exactly the tools groundcrew's own workflow depends on. Anything else stays an explicit `local.readOnlyDirs` line.

Codex is excluded: its `CODEX_HOME` is relocated to a staged writable home rather than read in place.

### Chains are followed to the end

Dotfiles layouts nest:

```
~/.claude/skills                             -> ~/.dot_files/config/claude/skills
~/.dot_files/config/claude/skills/cb-babysit -> ~/.dot_files/agents/skills/cb-babysit
```

Granting only the first hop is **worse than granting nothing**: the agent can *list* `~/.claude/skills` and sees every skill name, but reading any `SKILL.md` returns EPERM, so Claude Code reports the skill as merely unavailable and no permission error ever reaches the user. That failure mode cost real debugging time on this ticket — the natural hypotheses (wrong `--settings`, bad restart, agent misconfiguration) are all plausible and all wrong.

So every resolved directory is walked again for the links inside it, with a visited set (cycles terminate) and a depth bound (a launch is not a full-tree crawl). Only config entries of the agent dir are crawled: top-level symlinks whatever their name, plus the real `agents`, `commands`, `contexts`, `hooks`, `plugins`, `skills` subdirectories. `~/.claude/projects` and friends are runtime state — walking them is slow and would auto-grant stray transcript links.

### Safety guard

Targets resolving to `$HOME` (literal or realpath'd) or the filesystem root are refused, logged via `writeError` naming the path and pointing at `local.readOnlyDirs`, and **not descended into** — so a link to `/` never crawls the filesystem. A read-only grant of the home directory would expose `~/.ssh`, `~/.aws/credentials`, and every other repo to an agent with network egress.

The distinction is deliberate: `local.readOnlyDirs` is an explicit line in `crew.config.ts` that shows up in review, whereas symlink resolution is a silent inference at launch. `local: { readOnlyDirs: ["~"] }` still works — the guard requires saying it out loud. Failure modes are asymmetric and resolved in that direction: an over-cautious skip produces a legible EPERM naming the exact file, an over-eager grant produces no signal at all.

Every auto-grant is logged via `debug` (always tees to the log file, echoed with `--verbose`).

## Verification

Against the reporting machine's real home: **84 grants in 70ms**, including the two-hop `cb-babysit` skill that motivated the nested-chain fix.

25 unit tests in `sandboxSymlinkGrants.test.ts` — file-vs-directory link, all nine dotfiles entries, nested chain (partial-grant failure mode), chain ending at a file, symlink cycle, `$HOME` refusal (direct and through a chain), `/` refusal, broken link, duplicate targets, depth bound, runtime-state skip, symlinked roots, `~/.config/gh`, `~/.gitconfig` — plus 4 in `agentLaunch.test.ts` covering the `--add-dirs-ro` plumbing and the `none` runner.

`node --run verify` passes. Note it must be run with a clean `HOME` to match CI; a developer home with a deep real config tree can mask a coverage gap, which is how the first CI failure happened.

## Docs

`docs/configuration.md` gains a `Dotfiles-managed config (auto-resolved symlinks)` section recording the mechanism, the scope decision and its two rejected alternatives, the chain behavior, and the guard. The `local.readOnlyDirs` row links to it.

Closes atd-456

https://claude.ai/code/session_01Nu2DULU2uxfEUhF4MMHRfh


src/lib/sandboxSymlinkGrants.ts
```diff
@@ -0,0 +1,129 @@
...
+    const entryPath = path.join(input.dir, entry.name);
+    if (entry.isSymbolicLink()) {
+      const target = resolveRealPath(entryPath);
+      if (target !== undefined) {
+        yield target;
+      }
+      continue;
+    }
+    if (input.depth < MAX_WALK_DEPTH && entry.isDirectory()) {
+      yield* walkSymlinkTargets({ dir: entryPath, depth: input.depth + 1 });
```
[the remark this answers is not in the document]
bionboy: Already handled by https://github.com/ClipboardHealth/groundcrew/commit/1498561c6ab92e19fb3e86ff60953f371ae486ad. That commit adds a test asserting the depth bound directly (a link at the bound resolves, one past it does not) — same fix, no V8 ignore hint. The review ran against the PR's first commit; the branch coverage threshold has been green since. The bound itself has since moved to 6 in `sandboxSymlinkGrants.ts` to accommodate nested dotfiles chains, and the test was updated with it.

<sub>🤖 <code>cb-babysit:addressed v1 skill@1.0.8</code></sub>

bionboy:
## Review-body findings

**Already fixed**

- Depth-limit branch in `walkSymlinkTargets` uncovered, failing the 100% branch threshold — handled by https://github.com/ClipboardHealth/groundcrew/commit/1498561c6ab92e19fb3e86ff60953f371ae486ad, which asserts the bound directly rather than adding a V8 ignore hint. CI has been green since. The review ran against the PR's first commit (`a2742c2`); two commits have landed after it, including a rename of `agentConfigSymlinks.ts` to `sandboxSymlinkGrants.ts` and transitive symlink-chain resolution.

```
7ea6dcdbaaf20b8f
```

<sub>🤖 <code>cb-babysit:addressed v1 skill@1.0.8</code></sub>

bionboy:
**Review: blocking issue — the git failure this PR was dispatched to fix is still present.**

The resolver follows symlinks only, but `~/.gitconfig` on the reporting machine is a real file whose indirection is a git directive (`include.path = ~/.dot_files/.gitconfig`). `.gitconfig` is in `TOOL_CONFIG_PATHS`, `isSymbolicLink()` returns false, the walk `readdir`s a file and yields nothing. Verified by running `resolveSandboxSymlinkGrants` against the real home and launching a sandbox with **only** its 84 auto-grants: every git command still dies with `fatal: unable to access '/Users/luke/.dot_files/.gitconfig': Operation not permitted`.

Suggested fix: replace the `.gitconfig` entry with a `git config --list --show-origin` query and grant the origin files. That resolves `include.path` and `includeIf` for free, including conditional includes (`~/.gitconfig-clipboard`) that symlink-walking can never discover.

Three medium findings and two minor ones below. The symlink walker itself is correct — nested chains, file-vs-directory granting, loop protection, depth bound, and broken-link handling all check out, and the docs section is good.

<details>
<summary><b>Blocking — <code>include.path</code> indirection is not resolved (evidence)</b></summary>

`~/.gitconfig` (real file, not a symlink):

```ini
[include]
  path = ~/.dot_files/.gitconfig
[includeIf "hasconfig:remote.*.url:https://github.com/ClipboardHealth/**"]
  path = /Users/luke/.gitconfig-clipboard
```

Resolver output against that home:

```
gitconfig include target granted? false
cb-babysit skill granted?        true
gh config granted?               true
```

Sandbox launched with only the PR's auto-grants:

```
--- git config user.name ---
fatal: unable to access '/Users/luke/.dot_files/.gitconfig': Operation not permitted
--- auth to github (ls-remote) ---
(no output — failed)
```

Adding the include target to `--add-dirs-ro` makes both succeed, so the grant list is the only thing missing.

This fails the PR's own acceptance criterion: *"a symlinked `~/.config/gh/config.yml` resolves, so `gh` starts and the agent can open a PR without a manual `local.readOnlyDirs` entry."* `gh` config does resolve; git does not, and git breaks first.

`git config --list --show-origin` enumerates every file git actually loads, so it needs no inference about how the indirection was expressed. Worth considering the same shape for other tools later rather than guessing at their config layouts.

</details>

<details>
<summary><b>Medium — grant explosion: 74 of 84 grants are individual skill directories</b></summary>

`resolvedLinkTargets` calls `walkSymlinkTargets` without passing `descendNames`, so once the walk crosses into the dotfiles repo, descent into real subdirectories is unrestricted up to `MAX_WALK_DEPTH`. On the reporting machine `~/.claude/skills` → `~/.dot_files/config/claude/skills`, whose ~74 entries each link on to `~/.dot_files/agents/skills/<name>`, producing one grant apiece.

Measured: 84 grants, 4333 characters, 110 ms.

That works today, but the joined path list is composed into a shell command delivered through tmux, and it scales with the size of the user's dotfiles repo rather than with anything groundcrew bounds. Collapsing sibling grants that share a parent — grant `~/.dot_files/agents/skills` once instead of 74 children — would keep the list proportional to the number of *trees* involved rather than the number of files in them.

Counter-argument worth weighing: per-child grants are tighter, and the current behaviour is the more conservative one. If the explosion is considered acceptable, a cap with a loud log when it trips would at least bound the failure.

</details>

<details>
<summary><b>Medium — auto-grants are logged below the visibility threshold</b></summary>

Refusals go through `writeError`; grants go through `debug`, which is hidden unless `GROUNDCREW_VERBOSE` is set. So the sandbox silently gains 84 read-only paths with nothing said.

The originating ticket asked for the opposite, and for a specific reason: an over-cautious skip produces a legible EPERM naming the exact path, while an over-eager grant produces no signal at all. The asymmetry is the whole argument for the guard, and it only holds if grants are visible. A one-line summary at normal level (`auto-granted N read-only paths from ~/.claude, ~/.config/gh` with the full list behind `debug`) would satisfy it without noise.

</details>

<details>
<summary><b>Medium — <code>plugins</code> descent is unbounded below the first level</b></summary>

`plugins` is in `AGENT_CONFIG_DESCEND_NAMES`, and `~/.claude/plugins/cache` holds full cached plugin repositories. `descendNames` only constrains the first level, so everything under it is crawled to depth 6 on every launch.

110 ms total on the reporting machine, so this is a latency note rather than a defect — but it is per-launch cost scaling with a directory whose size nothing in groundcrew controls. Either exclude `cache`, or bound the descent under `plugins` more tightly than the generic depth limit.

</details>

<details>
<summary><b>Minor — guard omits <code>$HOME</code>'s parent, plus a garbled comment</b></summary>

`degenerateTargets` covers `/`, the filesystem root, and both literal and realpath-resolved `$HOME`. It does not cover `$HOME`'s parent (`/Users` on macOS, `/home` on Linux), where a grant would expose every user's home directory. Unlikely to occur, and cheap to add given the guard already exists.

In `agentLaunch.ts`, this comment does not parse:

> both sources are deduped and existence-filtered because safehouse rejects duplicate-free-but-absent `--add-dirs-ro` paths

Two separate reasons appear to have been spliced together. Safehouse rejects paths that do not exist; dedupe is for a different purpose.

</details>

<details>
<summary><b>What was verified, and how</b></summary>

- Read `sandboxSymlinkGrants.ts` in full, plus the `agentLaunch.ts` wiring and the docs diff.
- Ran `resolveSandboxSymlinkGrants({agent: "claude", homeDir: $HOME})` against a real dotfiles-managed home (three distinct indirection paths: `~/.claude/*` symlinks, a second hop from each skill link into `~/.dot_files/agents/skills`, and the gitconfig include).
- Launched `safehouse --add-dirs-ro=<only the auto-grants>` and exercised `git config`, `git ls-remote`, skill file reads, and `~/.gitconfig` reads inside the resulting sandbox.
- Confirmed skills and `gh` config are genuinely fixed by this PR: reading `~/.claude/skills/cb-babysit/SKILL.md` inside the sandbox succeeds, where it previously returned EPERM.
- CI on the head commit is green; the earlier `main` failure was a stale run.

Not verified: behaviour on Linux, and the `sdx` / `none` runner paths (unchanged by inspection, not executed).

</details>

bionboy:
## Conversation tab comments

All five findings addressed in https://github.com/ClipboardHealth/groundcrew/commit/cb776c53be55eb19f2e84acd74753ceb0e8d2b1d.

**Agree**

- **Blocking — `include.path` indirection not resolved.** Confirmed: `~/.gitconfig` is a real file, so the symlink walk finds nothing and git still dies on the include target. Replaced the `.gitconfig` entry in `TOOL_CONFIG_PATHS` with a `git config --list --show-origin --name-only -z` query run from the worktree, granting every file git reports loading. Against the reporting home that now grants `~/.gitconfig`, `~/.dot_files/.gitconfig`, `~/.gitconfig-clipboard` (the `includeIf` conditional, which symlink walking could never have found), and `/opt/homebrew/etc/gitconfig`. Best-effort: a git that cannot run leaves the pre-existing behavior rather than aborting the launch.
- **Auto-grants logged below the visibility threshold.** One `log()` line at normal level with the count and the roots; individual paths stay on `debug`. Nothing is granted silently now.
- **`plugins` descent unbounded.** Directories named `cache` or `node_modules` are never descended into, at any level.
- **Guard omits `$HOME`'s parent.** `/Users` and `/home` are now refused alongside `$HOME` and `/`.
- **Garbled comment in `agentLaunch.ts`.** Rewritten: absent paths are dropped because safehouse rejects them; dedupe is so an explicit entry and an auto-grant of the same path emit one flag.

**Agree in part**

- **Grant explosion (74 skill directories).** Took the bounding half, not the collapsing half. A target already inside a granted directory is now pruned (both as targets arrive and once at the end, since a parent tree can be granted after a file inside it), and the total is capped at 256 with a loud `writeError` when it trips — a truncated grant list means EPERM somewhere, so it must not be quiet. Deliberately kept per-child grants rather than collapsing 74 siblings into `~/.dot_files/agents/skills`: as you noted, per-child is the tighter grant, and widening to the shared parent would re-open whatever else lives in that tree. Current cost on the reporting home is 89 grants / 4503 chars / 88 ms, comfortably under the cap.

Verification: 35 unit tests in `sandboxSymlinkGrants.test.ts` (added: include-path origins, non-file origins, git failure, missing origin file, `$HOME` origin refusal, home-parent refusal, containment prune, `cache` skip, cap trip, summary log). `node --run verify` passes with a clean `HOME`. Re-probed against the real home: the gitconfig include target and the conditional include are both granted, and `cb-babysit`'s two-hop skill still resolves.

Not addressed: Linux and the `sdx`/`none` runner paths remain unexecuted, same as your review noted.

```
bc4be7e8ddf8859d
```

<sub>🤖 <code>cb-babysit:addressed v1 skill@1.0.8</code></sub>

bionboy:
**Re-review of `cb776c5` — blocking issue resolved, verified end to end on the reporting machine.** CI green.

All five findings addressed. The `git config --list --show-origin --name-only -z` approach works, and running it from the worktree also picks up the conditional `includeIf` file that symlink-walking could never have found. One residual note below, not blocking.

<details>
<summary><b>Verification (run against the real dotfiles-managed home)</b></summary>

Resolver output, at `cb776c5`:

```
grants: 89   chars: 4503   ms: 114
gitconfig include target granted?      true
gitconfig-clipboard (conditional)?     true
```

Sandbox launched with **only** the auto-grants — no `local.readOnlyDirs` entry — running the same probes that failed on the previous head:

```
--- read ~/.gitconfig ---      # This is Git's per-user configuration file.
--- git config user.name ---   Luke Floden
--- ls-remote ---              e5da68299c5d43a5cea0b62ff8fa0f2c7ebd096d  HEAD   (rc=0)
```

Previously `git config` aborted with `fatal: unable to access '/Users/luke/.dot_files/.gitconfig'` and `ls-remote` failed with it. Both now succeed, so the acceptance criterion the earlier head missed is met: an agent on a dotfiles-managed machine can run git and open a PR with no manual config.

The visible summary line is present and reads well:

```
Auto-granted 89 read-only sandbox path(s) resolved from .claude, .config/gh, .config/git
and git config; run with --verbose to list them.
```

</details>

<details>
<summary><b>Residual — the sibling-collapse comment overstates what the code does</b></summary>

`isCoveredByGrantedDir` drops a child when an ancestor is *already granted*. In the layout that motivated the finding, no ancestor is granted: the 74 skill links resolve into `~/.dot_files/agents/skills/<name>`, and `~/.dot_files/agents/skills` itself is never a grant target — only `~/.dot_files/config/claude/skills` (the intermediate hop) is. So the count went 84 → 89, not down.

That is defensible: per-child grants are tighter than granting the parent tree, and `MAX_AUTO_GRANTS` now bounds the worst case loudly. Reading this as an intentional choice of tightness over brevity, in which case nothing needs to change in the code.

The doc comment, though, claims something the code does not do:

> Keeps the list proportional to the number of trees involved rather than repeating a subtree already opened.

On this machine the list is proportional to the number of *skills*. Suggest narrowing the comment to what the helper actually guarantees — a child is not granted twice when its parent tree is already open — so the next reader does not trust a property that only holds when an ancestor happens to be granted.

</details>

<details>
<summary><b>Confirmed fixed, briefly</b></summary>

- **`include.path` indirection** — `gitConfigOriginPaths` asks git directly; `-z` and the `file:` prefix check are both handled correctly, and `runCommand` failure degrades to `[]` rather than aborting a launch.
- **`plugins` crawl** — `SKIPPED_DIR_NAMES` covers `cache` and `node_modules` wherever they appear. 114 ms on this machine, unchanged from before, so the bound costs nothing here.
- **Grant visibility** — `log()` summary at normal level, full list behind `--verbose`, and the cap warning via `writeError`. Matches what the ticket asked for.
- **`$HOME` parent** — `path.dirname(resolvedHome)` covers `/Users` and `/home`.
- **Garbled comment** in `agentLaunch.ts` — rewritten, now states the two reasons separately.

Still not verified by me: Linux behaviour, and the `sdx` / `none` runner paths (read, not executed).

</details>

src/lib/sandboxSymlinkGrants.ts
```diff
@@ -0,0 +1,397 @@
...
+    return [];
+  }
+  const origins = new Set<string>();
+  for (const record of (output ?? "").split("\0")) {
+    // Each record is "<origin>\n<name>"; only file origins name a readable path.
+    const [origin] = record.split("\n");
+    if (origin?.startsWith("file:") !== true) {
+      continue;
+    }
+    const originPath = path.resolve(input.homeDir, origin.slice("file:".length));
```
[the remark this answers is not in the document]
bionboy: Addressed in https://github.com/ClipboardHealth/groundcrew/commit/b7962417f693268d59e5469ae16249ea8e26d63d. Real escalation path, and the reachable version is worse than the repro suggests: the sandboxed agent has write access to its own worktree, so it could write `include.path` into `.git/config` and widen its *next* launch's grants to `~/.aws/credentials`.

`gitConfigOriginPaths` now passes `--show-scope` and grants only `system` and `global` origins; `local`, `worktree`, `command`, and `submodule` are dropped. Nothing is lost — repo-local config lives inside the worktree, which safehouse already grants. Regression test seeds `local` and `worktree` origins pointing at a credentials file and asserts an empty grant list, plus a companion test that `system` scope still resolves. Re-probed against a real dotfiles home: the `include.path` target and the `includeIf` conditional are still granted.

<sub>🤖 <code>cb-babysit:addressed v1 skill@1.0.8</code></sub>

bionboy:
## Review-body findings

**Agree** — fixed in https://github.com/ClipboardHealth/groundcrew/commit/b7962417f693268d59e5469ae16249ea8e26d63d

- **Critical: repo-local git config could widen the sandbox.** `gitConfigOriginPaths` granted every origin git reported, including `local` and `worktree` scope. The sandboxed agent can write its own `.git/config`, so an `include.path` there would name any host file and be auto-granted on the next launch. Now passes `--show-scope` and grants only `system` and `global`; `local`, `worktree`, `command`, and `submodule` are dropped. Repo-local config lives inside the already-granted worktree, so nothing is lost. Regression test covers a `local`/`worktree` include pointing at a credentials file.

## Conversation tab comments

**Agree** — fixed in https://github.com/ClipboardHealth/groundcrew/commit/b7962417f693268d59e5469ae16249ea8e26d63d

- **The `isCoveredByGrantedDir` comment overstates what the helper does.** Correct — on your machine the list is proportional to the number of skills, not trees, because `~/.dot_files/agents/skills` is never itself a grant target. Comment narrowed to what it actually guarantees (a child is not granted twice when its parent tree is already open) and it now points at `MAX_AUTO_GRANTS` as the thing that bounds the list. And yes — per-child over parent-collapse is the intentional choice, tightness over brevity.

Thanks for the end-to-end verification; `ls-remote` succeeding inside the sandbox with no `local.readOnlyDirs` entry is the check that actually closes the ticket.

```
6a018c7bafed5b40
35856528b1af981c
```

<sub>🤖 <code>cb-babysit:addressed v1 skill@1.0.8</code></sub>

bionboy:
**Re-review of `b796241` — scope filtering is correct and the hardening is a real catch.** `main` green; semgrep still running.

Dropping repo-local scope closes a genuine hole: `.git/config` is writable by the sandboxed agent and by repo-controlled hooks, so honoring an `include.path` there would have let a repository widen its own sandbox on the next launch. That is a sharper version of the guard the ticket asked for, and it costs nothing since the worktree is already granted.

One minor robustness note below. Nothing blocking.

<details>
<summary><b>Verification (real machine, at <code>b796241</code>)</b></summary>

Confirmed the wire format the new parser depends on — `--show-scope --show-origin --name-only -z` emits NUL-separated *triples*, not tab-joined records, so the field-at-a-time loop is reading the right shape:

```
system \0 file:/opt/homebrew/etc/gitconfig \0 credential.helper \0
global \0 file:/Users/luke/.gitconfig      \0 user.name         \0
```

Resolver output:

```
grants: 88                                  (was 89 at cb776c5)
dotfiles .gitconfig (global include)?  true
gitconfig-clipboard (conditional)?     true
worktree .git/config leaked?           false
```

So the one dropped grant is exactly the repo-local config, and both global-scope includes — plain and conditional — survive. The end-to-end sandbox result from the previous review is unaffected.

</details>

<details>
<summary><b>Minor — an unrecognized scope inherits the previous entry's trust</b></summary>

`trustedScope` is a running flag updated only when a field matches `TRUSTED_CONFIG_SCOPES` or `UNTRUSTED_CONFIG_SCOPES`. A scope token in neither set leaves the flag at its previous value, so if it follows a `global` entry, that unknown scope's origin is granted as though it were trusted.

Every scope git currently emits is covered (`system`, `global`, `local`, `worktree`, `command`; `submodule` is listed but not a scope git reports), so this is latent rather than live. It fails open, though, and in the direction the rest of the file is careful to fail closed.

Two options, either fine:

- default the flag to `false` on any field that is neither a known scope, an origin, nor a name
- parse positionally — the records are fixed triples, so `index % 3` identifies scope/origin/name exactly and no set membership is needed

The second also removes a smaller latent hazard: a config *key* that happened to equal a scope name would flip the flag today. `--name-only` keys always contain a dot, so it cannot currently happen, but positional parsing makes it structurally impossible rather than accidentally safe.

</details>

<details>
<summary><b>Comment fix confirmed</b></summary>

The `isCoveredByGrantedDir` doc now says what the code does — collapses only when an ancestor is granted, siblings still granted individually, `MAX_AUTO_GRANTS` is the actual bound. That was the residual note from the previous review; it is closed.

</details>

bionboy:
**Re-review of `aadbd3f` — no findings.** Behavior is unchanged on the reporting machine, the tests still pass, and the previous minor finding is closed.

A −152/+70 refactor with no test changes deserves scrutiny, so I checked the two substantive changes buried under the comment trimming rather than reading it as prose-only.

<details>
<summary><b>Scope parsing now fails closed (previous minor finding, resolved)</b></summary>

```js
if (!field.startsWith("file:")) {
  trustedScope = TRUSTED_CONFIG_SCOPES.has(field);
  continue;
}
```

Dropping `UNTRUSTED_CONFIG_SCOPES` in favour of assigning from the trusted set on every non-origin field is the better shape: an unrecognized scope now yields `false` instead of inheriting the previous entry's trust. That was the fail-open path I flagged, and it is gone without adding a second list to keep in sync.

Traced the field sequence to confirm nothing else flips the flag. Within a triple the name field (`user.name`) also fails the `file:` test and sets `trustedScope = false`, which is harmless — the next triple opens with its own scope. Non-file origins (`blob:`, `command line:`) land in the same branch and correctly evaluate to untrusted. `--name-only` guarantees no value field can impersonate a scope token.

</details>

<details>
<summary><b><code>isUnderGrantedDir</code> — equivalent, and the boundary case is right</b></summary>

The ancestor-walk became a prefix scan:

```js
[...grants].some((granted) => targetPath.startsWith(granted + path.sep))
```

Appending `path.sep` is what makes this safe: a grant of `.../skills` does not swallow a sibling `.../skills-extra`, which a bare `startsWith` would. Behaviour matches the old `dirname` loop for both file and directory grants.

It is O(n) per call inside an O(n) loop, so O(n²) overall, bounded by `MAX_AUTO_GRANTS` at 256 — about 65k string comparisons worst case. Measured 92 ms end to end on this machine versus 114 ms before, so this is noise, not a regression.

</details>

<details>
<summary><b>Verification</b></summary>

```
vitest sandboxSymlinkGrants.test.ts agentLaunch.test.ts → 63 passed
```

Resolver against the real dotfiles-managed home, at `aadbd3f`:

```
grants: 88                          (identical to b796241)
dotfiles .gitconfig?          true
conditional include?          true
repo-local .git/config leaked? false
cb-babysit skill?             true
```

On the comment trimming: the rationale that survives is the load-bearing part — partial-grant-worse-than-none, asymmetric failure modes, why the tool list is an allowlist, why local scope is dropped. The long-form version also still lives in the `docs/configuration.md` section added earlier in this PR, which this commit does not touch, so nothing is lost from the record.

</details>

src/lib/sandboxSymlinkGrants.ts
```diff
@@ -0,0 +1,351 @@
...
+}): Generator<ResolvedTarget> {
+  const targetPath = resolveRealPath(input.linkPath);
+  if (targetPath === undefined) {
+    return;
+  }
+  if (input.guard.has(targetPath)) {
+    yield { targetPath, refused: true };
+    return;
+  }
+  yield { targetPath, refused: false };
```
therockstorm: **[CRITICAL] Constrain inferred read grants**

This accepts every resolved target except broad roots; confine inferred targets to a host-authorized root and reject delimiter-bearing paths before serialization.

**Why it matters:** A symlink to `~/.aws/credentials` exposes those credentials to the next network-capable agent without explicit authorization.

src/lib/sandboxSymlinkGrants.ts
```diff
@@ -0,0 +1,351 @@
...
+import { debug, log, writeError } from "./util.ts";
+
+/**
+ * Where each agent keeps its config. Codex is absent on purpose: its
+ * `CODEX_HOME` is relocated into a staged, writable home rather than read in
+ * place (see `codexConfigRelocation.ts`).
+ */
+const AGENT_CONFIG_DIRS: Record<string, string> = {
+  claude: ".claude",
+};
```
therockstorm, line 18: **[MAJOR] Resolve Pi config symlinks**

ATD-456 requires equivalent config directories for supported agents, but this registry omits Pi's documented `~/.pi/agent`.

**Why it matters:** Dotfiles-managed Pi settings and credentials remain unreadable, preserving the manual workaround this ticket is meant to remove.

**Suggested fix:**

```suggestion
const AGENT_CONFIG_DIRS: Record<string, string> = {
  claude: ".claude",
  pi: ".pi/agent",
};
```

src/lib/agentLaunch.ts
```diff
@@ -108,7 +109,14 @@ export function composeAgentLaunch(input: {
           input.runner === "safehouse" ? input.safehouseEnableFeatures : undefined,
         // Safehouse rejects nonexistent --add-dirs-ro paths, so drop absent ones.
         safehouseAgentAddDirsReadOnly:
-          input.runner === "safehouse" ? (input.readOnlyDirs ?? []).filter(existsSync) : undefined,
+          input.runner === "safehouse"
```
therockstorm, line 113: **[MINOR] Avoid logging discarded grants**

Gate resolution on the same owned-wrap predicate; today it logs grants before `buildLaunchCommand` discards them for cmd-owned Safehouse commands.

**Why it matters:** Operators see “Auto-granted” while access remains denied, sending EPERM investigations in the wrong direction.

src/lib/sandboxSymlinkGrants.ts
```diff
@@ -0,0 +1,351 @@
...
+  const roots = [...(agentConfigDir === undefined ? [] : [agentConfigDir]), ...TOOL_CONFIG_PATHS];
+  const guard = degenerateTargets(input.homeDir);
+  const grants = new Set<string>();
+  let capped = false;
+
+  function addGrant(targetPath: string, source: string): void {
+    if (grants.has(targetPath) || isUnderGrantedDir(targetPath, grants)) {
+      return;
+    }
+    if (grants.size >= MAX_AUTO_GRANTS) {
```
therockstorm, line 148: **[MINOR] Canonicalize before enforcing the cap**

Covered descendants are removed only after the 256-path cap and log; canonicalize against auto and explicit parent grants before both.

**Why it matters:** Redundant paths can exhaust the cap and omit a required config target even when the final list is small.

therockstorm (commented):
_These comments were generated by @therockstorm using Claude Code._

This PR fixes symlinked Claude/tool configuration under Safehouse, but should not ship as-is: recursive traversal lets installed config content expose arbitrary host files, and Pi support required by ATD-456 is missing.

Spec verdict: missing.

**Apply all comments at once** — paste this into Claude Code on a checkout of this branch:

```
Fetch the review by @therockstorm on PR https://github.com/ClipboardHealth/groundcrew/pull/388 that contains this exact prompt. For every inline comment in that review, address the issue: when the comment includes a `suggestion` block, apply it verbatim; otherwise implement an equivalent fix that satisfies the comment's "Why it matters" rationale. After resolving each thread, post a reply on that thread with a one-line summary of what you changed. When all comments are handled, run the project's tests, commit the changes with a message that references the review, and report back any threads you could not resolve and why.
```

src/lib/sandboxSymlinkGrants.ts
bionboy: Addressed in https://github.com/ClipboardHealth/groundcrew/commit/c51b189, both halves.

**Credential stores refused.** Real reachable path, and worth naming precisely: the agent can write its own config dir, so it could plant `~/.claude/settings.json -> ~/.aws/credentials` and have the *next* launch grant it. Same escalation shape as the repo-local git scope fix earlier in this PR. Added `REFUSED_TARGET_DIRS` — `.ssh`, `.aws`, `.gnupg`, `.kube`, `.docker`, `.config/gcloud`, `.config/op`, `Library/Keychains` — matched against the resolved target and its ancestors, refused with a log line pointing at `local.readOnlyDirs`, and never descended into.

I did not implement the 'confine to a host-authorized root' half as stated, because on a dotfiles machine the authorized root would have to be `$HOME` (that is where both the dotfiles repo and `~/.aws` live), so it would not exclude the thing you are worried about. The denylist excludes it directly. Calling that out as a deliberate divergence: it is a barrier on known-valuable paths, not a proof, and I would rather it read that way in the code than look like a boundary it is not.

**Delimiter-bearing paths rejected.** Confirmed against `safehousePathListFlag` in `launchCommand.ts:869` — `paths.join(":")` inside `shellSingleQuote`, so quoting is safe but a `:` in a path silently becomes two grants. Any target containing `:` is now refused with its own message. Tests cover both refusals.

<sub>🤖 <code>cb-babysit:addressed v1 skill@1.0.8</code></sub>

src/lib/sandboxSymlinkGrants.ts
bionboy: Addressed in https://github.com/ClipboardHealth/groundcrew/commit/c51b189. Took the suggestion as written; verified against this repo's own docs first — `docs/credentials.md:31` documents `~/.pi/agent/auth.json` and `docs/runners.md:66` names `~/.pi/agent` as the state dir Safehouse must open.

One caveat I left in a code comment: `PI_CODING_AGENT_DIR` relocates that base directory and is not followed here, since per `docs/runners.md` a custom value needs its own Safehouse env and writable-path grants anyway. Also worth flagging that the grant is read-only, so a dotfiles-symlinked `auth.json` becomes readable but a token *refresh* through the link would still fail — strictly better than the current EPERM-on-read, and the same shape as claude's settings. Codex stays excluded (relocated `CODEX_HOME`), and I did not add cursor since nothing in the repo documents its config dir — guessing at layouts is what the `~/.config` allowlist exists to avoid.

<sub>🤖 <code>cb-babysit:addressed v1 skill@1.0.8</code></sub>

src/lib/agentLaunch.ts
bionboy: Addressed in https://github.com/ClipboardHealth/groundcrew/commit/c51b189. Confirmed: `shouldWrapWithSafehouse` (`launchCommand.ts:590`) drops every groundcrew-composed flag when `cmd` matches `/^safehouse(\s|$)/`, so the resolver was doing the work, logging 'Auto-granted N read-only sandbox path(s)', and having the flags discarded.

Extracted that predicate as `groundcrewOwnsSafehouseWrap(cmd)` and gated resolution on it in `resolveSafehouseAddDirsReadOnly`, so a cmd-owned wrap resolves nothing and says nothing. Test asserts a `safehouse -- claude …` cmd produces no grant for a symlinked `~/.claude/skills`. Good catch on the failure mode — a log line claiming access that was denied is worse than silence.

<sub>🤖 <code>cb-babysit:addressed v1 skill@1.0.8</code></sub>

src/lib/sandboxSymlinkGrants.ts
bionboy: Addressed in https://github.com/ClipboardHealth/groundcrew/commit/c51b189. Canonicalization now happens on insert, before the cap and before the log, in two directions:

- A target already covered by an existing grant **or by a `local.readOnlyDirs` entry** is skipped — the explicit list is passed in as `explicitDirs` purely for containment, so an explicit entry no longer costs an auto-grant slot for everything under it.
- When a newly granted tree subsumes grants added earlier, those entries are deleted, returning their slots. That was the ordering hole: children resolved before their parent counted permanently against the 256, so a required target could be dropped while the final list was small. The end-of-function prune is gone, since the set is now canonical as it is built.

Tests cover both: an explicit parent suppressing an auto-grant, and three children granted before their parent collapsing to one entry.

<sub>🤖 <code>cb-babysit:addressed v1 skill@1.0.8</code></sub>

bionboy:
## Conversation tab comments

**Already fixed**

- Re-review of `b796241` — the minor robustness note (unrecognized scope inheriting the previous entry's trust) was closed by the `aadbd3f` refactor, which replaced the two-set scope check with `trustedScope = TRUSTED_CONFIG_SCOPES.has(field)` on every non-origin field. Fails closed now, as your follow-up confirmed.
- Re-review of `aadbd3f` — no findings; nothing to act on. Thanks for tracing the field sequence rather than taking the refactor as prose-only.

Both are superseded by [`c51b189`](https://github.com/ClipboardHealth/groundcrew/commit/c51b189), which addresses the four findings from @therockstorm's review — credential-store and delimiter refusals, Pi's `~/.pi/agent`, gating resolution on the owned-wrap predicate, and canonicalizing before the cap. Replies are on the individual threads.

```
42c29ae68d0cb464
6b0245559597a1d2
```

<sub>🤖 <code>cb-babysit:addressed v1 skill@1.0.8</code></sub>

bionboy:
**Closing this one out.** Initially thought handling symlinks would be straight forward.
Didn't use a draft-PR because I thought review bot would chime in; _apologies for any noise_!

therockstorm:
@bionboy Can you dd the resolved dotfiles root or target roots to `local.readOnlyDirs`. Groundcrew forwards those paths to Safehouse as `--add-dirs-ro`, so the linked Claude, Pi, `gh`, and Git configuration works without any new Groundcrew behavior.

Automatic discovery introduces a security problem. A legitimate link into `~/dotfiles` and an agent-planted link to `~/.netrc` look identical when the agent can write the directory containing the link. Adding more credential paths to the denylist only moves the hole.

Safehouse is the right owner for sandbox solutions. Implementing this in Groundcrew duplicates that knowledge and makes us maintain agent-specific resource directories, recursive traversal, Git includes, canonicalization, caps, and credential policy.

PR #390, opened 2026-09-04
Carry decisions across the per-task worktrees

drexthealpha:
*"One git worktree per task, sandboxed by default"* - `src/lib/worktrees.create.ts` and `worktreeRunState.ts` handle the isolation, and `worktreeRunState` is per-worktree by design. Nothing crosses between them, so the reason a task was done a particular way is lost to the next task that touches the same code.

**What's in the PR**

- `.knos/decisions.md` - a record shared by every worktree of the repo, keyed on `git rev-parse --git-common-dir` rather than `--show-toplevel`; the latter returns the worktree root, which is exactly what makes per-worktree state per-worktree. Seeded from this repo's own rules, each line citing its source.
- `.mcp.json` - one server, three tools, stdio.

It does not touch `worktreeRunState`. Run state should stay per-worktree; decisions should not.

**Disclosure:** I wrote Knos. It is a local MCP server - three tools, no ports, no network, no account, no model download. The record is seeded from your own docs rather than mine so it is checkable rather than promotional. Close it if a third-party entry isn't wanted; the `.knos/decisions.md` half needs no install and works as plain markdown either way.

### DIFF
--- a/.knos/decisions.md
+++ b/.knos/decisions.md
@@ -2,6 +2,14 @@
 
 <!-- Written by `knos export`. Commit this file. -->
 
+<!--
+Reading this file needs nothing installed: it is plain markdown, and a fresh
+clone picks it up as-is. The live claim/withhold server is a separate, optional
+step - `pip install knos` (Python 3.10+), which the MCP entry launches as
+`python -m knos.mcp`. Without it, everything below still reads normally.
+-->
+
+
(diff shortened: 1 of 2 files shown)


drexthealpha:
Closing this as not a fit for this repo's current contribution patterns. Thanks for the review time.

